PR #10, opened 2023-01-31
Add config page to broker details

AjayJagan:
Adds config page to the broker details

PR #11, opened 2023-02-01
Add husky precommit

AjayJagan:
Adding husky pre-commit hook to improve maintenance of code by using prettier and eslint only on staged files.


RoddieKieley (commented):
This by itself worked ok for me, although between this and the branch from PR #10 it appears there is some minor cleanup to do related to src/utils/utils.ts.

ajaypratap003 (approved):
Looks good.

PR #13, opened 2023-02-09
i18n initial setup

ajaypratap003:


RoddieKieley (approved):
I grabbed the branch and locally ran and tested it without issue. I also locally merged it into main, also without issue. The text looked correct still even with the indirection to the i18n setup. 

Functionally at least this appears good to me.

PR #22, opened 2023-03-31
Build components to display clients data

Msarawan:
Clients Page :

Screenshot from 2023-03-31 20-35-07


RoddieKieley (approved):
I ran this branch up here locally with no problems. Getting the display structure in place with boilerplate date is a good start to actually being able to display real clients data. Will merge.

PR #78, opened 2023-08-03
App folder restructuring

Msarawan:
Description:
- Modified App folder restructuring and verified everything should work as expected after this structural change.


ajaypratap003 (changes requested):
Folder name should be in small case as these are modules i.e. clients, configuration, queues, topics etc. And another issue I noticed on local that left side nav design doesn't look ok. Please verify it on your local before merging PR. Rest looks fine.

PR #76, opened 2023-08-03
Add test case for BrokerRow component

Msarawan:
- Added test cases for BrokerRow.tsx component

PR #81, opened 2023-08-23
Add unit test for BrokersList component

Msarawan:
Added unit test for BrokerList component


lavocatt:
@Msarawan it looks to me that this PR has stalled, can you review it and rebase it if it is still relevant?

lavocatt (changes requested):
Can you add a test also when there's no elements to display and there's no error at the same time please?

Msarawan:
closed as completed via PR [#76](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/pull/76)

PR #89, opened 2023-09-14
Unit tests for ChartPopover component

Msarawan:


src/metrics/components/ChartPopover/ChartPopover.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { render, screen, waitForI18n } from '../../../test-utils';
+import { ChartPopover } from '../ChartPopover/ChartPopover';
+
+jest.mock('../../../i18n', () => ({
+  useTranslation: () => ({
```
ajaypratap003, line 5: Why do we need to mock translation and adding condition for key. 
we can check translation key `chart_popover_icon_screenreader_text`
 
 ```
 expect(
      screen.getByLabelText('chart_popover_icon_screenreader_text'),
    ).toBeInTheDocument();
  ```

Msarawan: > Why do we need to mock translation and adding condition for key. we can check translation key `chart_popover_icon_screenreader_text`
> 
> ```
> expect(
>      screen.getByLabelText('chart_popover_icon_screenreader_text'),
>    ).toBeInTheDocument();
> ```

Yes Ajay initially i tried by checking the key itself that time also the test is passed .as in my first commit you can see the changes. i just checked the other way we  can apply so that instead of key we can check test case for value itself so that i modified this approach.

ajaypratap003 (approved):

PR #90, opened 2023-09-14
Unit tests for ChartSkeletonLoader component

Msarawan:


src/metrics/components/ChartSkeletonLoader/ChartSkeletonLoader.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { render, waitForI18n } from '../../../test-utils';
+import { ChartSkeletonLoader } from '../ChartSkeletonLoader/ChartSkeletonLoader';
+
+jest.mock('../../../i18n', () => ({
```
ajaypratap003, line 4: This is not right way to get the translation value. We will fix it later. You can check translation key instead. Don't need to mock translation. 

expect(comp.getByText('skeleton_loader_screenreader_text')).toBeInTheDocument();

Msarawan: > This is not right way to get the translation value. We will fix it later. You can check translation key instead. Don't need to mock translation.
> 
> expect(comp.getByText('skeleton_loader_screenreader_text')).toBeInTheDocument();

Yes Ajay, here also i just tried the trial and check method for test case by adding mocking translation. i will modify it to our common approach by expecting the test on key .

ajaypratap003 (approved):

PR #94, opened 2023-09-21
Unit tests for MetricsLayout component

Msarawan:


src/metrics/components/MetricsLayout/MetricsLayout.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { render, waitForI18n } from '../../../test-utils';
+import { MetricsLayout } from '../MetricsLayout/MetricsLayout';
+jest.mock('../../utils', () => {
```
ajaypratap003, line 3: Why do we need to mock `utils` ?

Msarawan: Actually instead of using real implementation of utils i thought to mock the utils folder so that it not uses actual implementation of utils.

Msarawan: if we din't used mocked utils then also it will executes the tests according to our conditions in the component. for that we need to directly import the MetricsType from utils directly in our test file.

lavocatt:
@Msarawan it looks to me that this PR has stalled, can you review it and rebase it if it is still relevant?

lavocatt (approved):

PR #96, opened 2023-09-22
Fix Table alignment issue

Msarawan:
Screenshot from 2023-09-22 14-56-41


Msarawan:
Fixed table-alignment issue. now the table data is displaying correctly.


> Screenshot from 2023-09-22 14-56-41

ajaypratap003 (approved):
Looks good. Make sure, changes should not break charts layout.

PR #98, opened 2023-12-01
Fix plugin loading issue in the openshift-console

Msarawan:


ajaypratap003 (approved):
Changes look fine. We need to look into console image latest tag. it unblock us for local development.

PR #104, opened 2024-01-25
Deploy plugin with ssl enabled

Msarawan:
Description:

Fixed deploying plugin with ssl enable. now the plugin is loading in the openshift dev-cluster.

Screenshot from 2024-01-25 18-13-48

Screenshot from 2024-01-25 14-53-55

### DIFF
--- a/http-server.sh
+++ b/http-server.sh
@@ -7,4 +7,4 @@ shift
 SERVER_OPTS="$@"
 
 echo "starting server with options: $SERVER_OPTS"
-./node_modules/.bin/http-server $PUBLIC_PATH -c-1 $SERVER_OPTS
+./node_modules/.bin/http-server $PUBLIC_PATH -c-1 --cors $SERVER_OPTS
--- a/http-server.sh
+++ b/http-server.sh
@@ -6,4 +6,5 @@ PUBLIC_PATH="$1"
 shift
 SERVER_OPTS="$@"
 
-./node_modules/.bin/http-server $PUBLIC_PATH -p 8001 -c-1 --cors $SERVER_OPTS
+echo "starting server with options: $SERVER_OPTS"
+./node_modules/.bin/http-server $PUBLIC_PATH -c-1 $SERVER_OPTS
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -26,7 +26,7 @@ spec:
     spec:
       containers:
         - name: activemq-artemis-self-provisioning-plugin
-          image: quay.io/artemiscloud/activemq-artemis-self-provisioning-plugin:dev.latest
+          image: quay.io/hgao/activemq-artemis-self-provisioning-plugin:1.0.0-con-latest-ssl
           ports:
             - containerPort: 9443
               protocol: TCP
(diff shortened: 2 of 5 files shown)


webpack.config.ts
```diff
@@ -3,6 +3,7 @@
 import { Configuration as WebpackConfiguration } from 'webpack';
 import { Configuration as WebpackDevServerConfiguration } from 'webpack-dev-server';
 import * as path from 'path';
+//import fs from 'fs';
```
RoddieKieley, line 6: Was it intentional to leave the comment in for future reference?

http-server.sh
```diff
@@ -6,4 +6,5 @@ PUBLIC_PATH="$1"
 shift
 SERVER_OPTS="$@"
 
-./node_modules/.bin/http-server $PUBLIC_PATH -p 8001 -c-1 --cors $SERVER_OPTS
\ No newline at end of file
+echo "starting server with options: $SERVER_OPTS"
+./node_modules/.bin/http-server $PUBLIC_PATH -c-1 $SERVER_OPTS
```
RoddieKieley, line 10: Was the removal of --cors intentional here? It's not added back in the ENTRYPOINT above.

manifest.yaml
```diff
@@ -66,35 +58,10 @@ spec:
...
-        root                /usr/share/nginx/html;
-      }
-    }
----
-apiVersion: v1
 kind: Service
 metadata:
   annotations:
-    service.alpha.openshift.io/serving-cert-secret-name: plugin-serving-cert
+    service.beta.openshift.io/serving-cert-secret-name: plugin-serving-cert
```
RoddieKieley, line 64: Are we eliminating compatibility with any older versions by moving from alpha to beta here? If so it's a good item to to point out via the README.md and other docs.

webpack.config.ts
```diff
@@ -58,7 +59,12 @@ const config: Configuration = {
   },
   devServer: {
     static: './dist',
-    port: 9001,
+    port: 9443,
```
RoddieKieley, line 62: If we were to change the port passed in the SERVER_OPTS would we also need to update here?

webpack.config.ts
```diff
@@ -58,7 +59,12 @@ const config: Configuration = {
...
+    https: true,
+    // {
+    //  key: fs.readFileSync('/var/serving-cert/plugin-serving-cert/key.pem'),
+    //  cert: fs.readFileSync('/var/serving-cert/plugin-serving-cert/cert.pem'),
+    //},
```
RoddieKieley, line 67: Should these be removed?

RoddieKieley (commented):
A few questions and items to consider before merge

webpack.config.ts
Msarawan: Nope...i will update the PR by removing the comment lines.

webpack.config.ts
Msarawan: yes!

webpack.config.ts
Msarawan: yes! i will update the PR by removing the comment lines.

http-server.sh
Msarawan: nope! i will update it back.

ajaypratap003 (approved):

PR #106, opened 2024-01-30
Add x86_64 to fix openshift origin image arch issue for macbook machine

ajaypratap003:
Add x86_64 to fix openshift orgin image arch issue for macbook machine


Msarawan (approved):
Tested on Fedora OS without getting any issues. Changes work fine!

PR #108, opened 2024-02-16
Fix Plugin loading and routing issues with the latest tag for local development

Msarawan:
Resolved plugin loading and routing issues with the latest tag for local development.

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -11,7 +11,7 @@
     "type": "console.page/route",
     "properties": {
       "exact": true,
-      "path": ["/k8s/ns/:ns/add-broker", "/k8s/all-namespaces/add-broker"],
+      "path": ["/k8s/ns/:ns/add-broker"],
       "component": { "$codeRef": "AddBrokerPage" }
     }
   },
@@ -19,10 +19,7 @@
     "type": "console.page/route",
     "properties": {
       "exact": true,
-      "path": [
-        "/k8s/ns/:ns/edit-broker/:name",
-        "/k8s/all-namespaces/edit-broker/:name"
-      ],
+      "path": ["/k8s/ns/:ns/edit-broker/:name", "/k8s/all-namespaces/edit-broker/:name"],
       "component": { "$codeRef": "UpdateBrokerPage" }
     }
   },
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "react-dom": "17.0.2",
     "react-router": "5.3.3",
     "react-router-dom": "5.3.0",
+    "react-router-dom-v5-compat": "^6.22.0",
```
AjayJagan, line 39: So if I understand this right, this library is like an intermediate migration kit.
In that case this will be a temporary solution right?
I mean, after some point, we need to start using this: https://www.npmjs.com/package/react-router-dom

Msarawan: Yeah, that's correct. When i import useParams from `react-router-dom`, there are namespace readability issues. However, i noticed in the dynamic-plugin-sdk code, they imported useParams from `react-router-dom-v5-compat` dependency, which resolves the problem. So, i thought of switch to that dependency to fix the issue with our plugin.

However, I noticed that someone has already created this issue with useParams in the ocp-bugs: 
https://issues.redhat.com/browse/OCPBUGS-29526

AjayJagan: great! thanks for the clarification

AjayJagan (approved):

console-extensions.json
```diff
@@ -11,7 +11,7 @@
     "type": "console.page/route",
     "properties": {
       "exact": true,
-      "path": ["/k8s/ns/:ns/add-broker", "/k8s/all-namespaces/add-broker"],
```
ajaypratap003, line 14: Why are you removing route path for all-namespaces i.e. `"/k8s/all-namespaces/add-broker"` ? We need it.

package.json
ajaypratap003: dynamic-sdk team has started using  lib `react-router-dom-v5-compat`  instead `react-router`. As we are dependent on dynamic-sdk so let's use same library. keep watch what libs dynamic-sdk team is using. We will have to change this lib in case dynamic-sdk team will change it otherwise continue.

ajaypratap003 (commented):
Added a comment about route path for all-namespaces. Rest looks good.

gaohoward (approved):

ajaypratap003 (approved):

Issue #121, opened 2024-04-17
eslint: add the of hook & eqeqeq


lavocatt:
The `eslint` configuration would  benefit from having the `react-hooks` plugin added to it.
That would prevent a contributor from making a mistake regarding hooks see https://legacy.reactjs.org/docs/hooks-rules.html

Also I propose to add `eqeqeq: error` as a rule for the linter. This is mainly to not forget that the triple equality is the one to use. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

PR #122, opened 2024-04-18
[#121] harden the eslint config

lavocatt:
This commit adds:
* the react-hooks plugin to the configuration
* the eqeqeq to the configuration

Also two issues regarding the react-hooks were treated:

Normally it is impossible to call a hook in a conditional branch (if,
loops, ...). That is because it would mess with how React stores it's
data. https://react.dev/reference/rules/rules-of-hooks However in this
specific configuration, the number determining how many loops are
going to occur gets defined by the number of pods within the deployment.
This means this number is guaranteed to stay the same while the page is
displayed, as long as there is no support for the hot reloading of the
deployment details.

/!\ I ran the linter with `yarn lint` and some files got modified along the way.

### DIFF
--- a/src/metrics/hooks/useFetchCpuUsageMetrics.tsx
+++ b/src/metrics/hooks/useFetchCpuUsageMetrics.tsx
@@ -29,6 +29,19 @@ export const useFetchCpuUsageMetrics = (
   return ({ name, namespace, span, samples, endTime, timeout, delay }) => {
     [...Array(n)].map((_, i) => {
       const query = cpuUsageQuery(name, namespace, i);
+      // Normally it is impossible to call a hook in a conditional branch (if,
+      // loops). That is because it would mess with how React stores it's data.
+      // https://react.dev/reference/rules/rules-of-hooks
+      //
+      // However in this specific configuration, the number `n` determining how
+      // many loops are going to occur gets defined by the number of pods within
+      // the deployment. This means this number is guaranteed to stay the same
+      // while the page is up and running, as long as there is no support for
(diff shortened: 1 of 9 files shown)


src/metrics/hooks/useFetchCpuUsageMetrics.tsx
```diff
@@ -29,6 +29,19 @@ export const useFetchCpuUsageMetrics = (
...
+      //
+      // However in this specific configuration, the number `n` determining how
+      // many loops are going to occur gets defined by the number of pods within
+      // the deployment. This means this number is guaranteed to stay the same
+      // while the page is up and running, as long as there is no support for
+      // hot reloading.
+      //
+      // That said, it means that it is therefore safe to use this hook in this
+      // loop.
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
Msarawan, line 44: @lavocatt as far i know disabling ESLint warnings with inline comments to fix issues is not the recommended approach in the codebase. Instead, we strive to maintain consistency and manage ESLint rules globally through our ESLint configuration file (eslintrc.yaml).

Msarawan:
Hello @lavocatt , could you please rebase and Update this PR.

src/metrics/hooks/useFetchCpuUsageMetrics.tsx
Msarawan: yes, it's makes sense to me. and i agree with point.

Msarawan (approved):
Changes LGTM.

PR #125, opened 2024-04-22
api server implementation

gaohoward:
This is the api-server implementation
for accesing jolokia endpoints of brokers.

Please see api.md for details.


Dockerfile
```diff
@@ -25,19 +25,43 @@ RUN yarn install  --network-timeout 1000000
...
+RUN npm install cors
+RUN npm install express
+RUN npm install express-openapi-validator
+RUN npm install swagger-routes-express
+RUN npm install typescript
+RUN npm install validator
+RUN npm install yamljs
+RUN npm install base-64
+RUN npm install jsonwebtoken
+RUN npm install dotenv
```
AjayJagan, line 52: just a small suggestion here. If I understand correctly, the `RUN` creates new layers. So can it be possible to use `&&` and combine the commands?

lavocatt (commented):
I think this a very good first step! I've got a few comments on how to better integrate with React. I'll summarize the main points there. I'll also add a few other points later on.

If I understand the workflow correctly, what we want to achieve is to call the function `LoginJolokia` in two situation:
* when the component is rendering the first time and to obtain a token.
This is achieved by having a main component  `LoginHandler` at the top level of the `BorkerDetails` component. When the `BrokerDetails` are rendered, the `LoginHandler` is rendered and triggers a login if necessary.
* when the user interacts with the form and wants to send a query to the backend.

I will focus on the first useCase first. 

I think the workflow is good but there are a few issues in how it is implemented:

* The `LoginHandler` is a function component. In the React terminology, this means that the function should return a JSX to be displayed on screen. However in our case, the LoginHandler returns null. This isn't necessarily an  issue but, it means that we might head in the wrong direction.

* `LoginJolokia` is a function that will trigger a `fecth` to an api. This means that we ware stepping oustide of React while executing the `fecth` and this can have side effects. We have to use a `useEffect` to call external apis.

* The login process is storing some data in the browser session, and there are some methods to know if a login is necessary or not depending on the presence of some data in the cache. I think that storing the data in a React state is preferable. This way we can use React to know if the value needs to be computed or not.

## proposal using useEffect

### why using useEffect

The core point of the review is to have the call to the backend api in a `useEffect` hook. This important because effects don't block the rendering and are scheduled after the rendering cycle. If for any reason the call to the jolokia login page would take longer than expected, than the UI could suffer from stuttering/blockage.

see:
* https://react.dev/learn/synchronizing-with-effects
* https://react.dev/reference/react/useEffect
* https://react.dev/learn/you-might-not-need-an-effect
* https://www.reddit.com/r/reactjs/comments/mgm8o3/why_is_useeffect_used_for_data_fetching/

### pseudo code with the effect:

```
const [token, setToken] = useState();
const [needLogin, setNeedLogin] = useState(true);
useEffect(() => {
  if (!token && needLogin){
    token, err = jolokiaLogin()
    if(!err){
      setToken(loginJolokia())
    } else {
      //trigger an error popup
    }
   setNeedLogin(false);
  }
}, [token,  needLogin])
```

This piece of code can live in the `BrokerDetails` and will get executed whenever the dependency array gets updated.

If you want, I can work on a proposal with this approach for your PR. Don't hesitate to ping me on slack or to organize a meeting to further discuss this.

Dockerfile
gaohoward: Thanks Ajay, didn't know that. I think I can do it using && or if I can use a package.json ?

AjayJagan: yea I guess we can use it both ways 😄 . either in the docker file or as a command in package.json like `install-deps`

api-server/api/controllers/api_impl.ts
```diff
@@ -0,0 +1,222 @@
...
+import {
+  ArtemisJolokia,
+  JolokiaExecResponse,
+  JolokiaObjectDetailsType,
+  JolokiaReadResponse,
+} from '../apiutil/artemis_jolokia';
+import { API_SUMMARY } from '../../utils/server';
+
+// get the broker mbean from jolokia
+export function getBrokers(_: express.Request, res: express.Response): void {
```
Msarawan, line 11: Could you please update the code to follow the ES6 syntax, as we follow in the project's codebase. (i.e)

```
export const getQueues = (req: express.Request, res: express.Response): void => {
   // rest of your code here
}
```

src/utils/api-server.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+interface LoginProp {
+  brokerDetail: K8sResourceKind;
+}
+
+export const LoginHandler: FC<LoginProp> = ({ brokerDetail }) => {
+  const brokerRoutes = useContext(RouteContext);
+
+  //none | ongoing | fail | ok
+  const [loginState, setLoginState] = useState('none');
```
lavocatt, line 377: It's possible to define as a type the set of valid states: `const [loginState, setLoginState] = useState<'none' | 'ongoing' | 'fail' | 'ok'>('none');`

lavocatt:
I'm wondering if you would be ok to split the PR in multiple commits? I think there's place for one commit for the api server, one for the frontend access to it and one for the changes to the webpack configuration. It would make it easier to follow the changes.

Dockerfile
gaohoward: @AjayJagan I found a easy way to do it. i.e. npm can install multiple package in one command, for example
npm install package1 package 2...

api-server/api/controllers/api_impl.ts
gaohoward: ok will do. Thanks!

gaohoward:
> I'm wondering if you would be ok to split the PR in multiple commits? I think there's place for one commit for the api server, one for the frontend access to it and one for the changes to the webpack configuration. It would make it easier to follow the changes.

It makes sense. I'll see what I can do (at least the api server part can be one commit. the other changes to the front end and webpack changes may be in one or two, as I don't think I made much changes to the webpack)

gaohoward:
> I think this a very good first step! I've got a few comments on how to better integrate with React. I'll summarize the main points there. I'll also add a few other points later on.
> 
> If I understand the workflow correctly, what we want to achieve is to call the function `LoginJolokia` in two situation:
> 
>     * when the component is rendering the first time to obtain a token.
>       This is achieved by having a main component  `LoginHandler` at the top level of the `BorkerDetails` component. When the `BrokerDetails` are rendered, the `LoginHandler` is rendered and triggers a login if necessary.
> 
>     * when the user interacts with the form and wants to send a query to the backend.
> 
> 
> I will focus on the first useCase first.
> 
> I think the workflow is good but there are a few issues in how it is implemented:
> 
>     * The `LoginHandler` is a function component. In the React terminology, this means that the function should return a JSX to be displayed on screen. However in our case, the LoginHandler returns null. This isn't necessarily an  issue but, it means that we might head in the wrong direction.
> 
>     * `LoginJolokia` is a function that will trigger a `fecth` to an api. This means that we ware stepping oustide of React while executing the `fecth` and this can have side effects. We have to use a `useEffect` to call external apis.
> 
>     * The login process is storing some data in the browser session, and there are some methods to know if a login is necessary or not depending on the presence of some data in the cache. I think that storing the data in a React state is preferable. This way we can use React to know if the value needs to be computed or not.
> 
> 
> ## proposal using useEffect
> ### why using useEffect
> 
> The core point of the review is to have the call to the backend api in a `useEffect` hook. This important because effects don't block the rendering and are scheduled after the rendering cycle. If for any reason the call to the jolokia login page would take longer than expected, then the UI could suffer from stuttering/blockage.
> 
> Another that is important in React is that component must be pure. Having to step out of the component for a request during the render of the component makes it impure therefore it has to be avoided.
> 
> see:
> 
>     * https://react.dev/learn/keeping-components-pure
> 
>     * https://react.dev/learn/synchronizing-with-effects
> 
>     * https://react.dev/reference/react/useEffect
> 
>     * https://react.dev/learn/you-might-not-need-an-effect
> 
>     * https://www.reddit.com/r/reactjs/comments/mgm8o3/why_is_useeffect_used_for_data_fetching/
> 
> 
> ### pseudo code with the effect:
> 
> ```
> const [token, setToken] = useState();
> const [needLogin, setNeedLogin] = useState(true);
> useEffect(() => {
>   if (!token && needLogin){
>     token, err = jolokiaLogin()
>     if(!err){
>       setToken(token)
>     } else {
>       //trigger an error popup
>     }
>    setNeedLogin(false);
>   }
> }, [token,  needLogin])
> ```
> 
> This piece of code can live in the `BrokerDetails` and will get executed whenever the dependency array gets updated.
> 
> If you want, I can work on a proposal with this approach for your PR. Don't hesitate to ping me on slack or to organize a meeting to further discuss this.

Thanks @lavocatt for the advice. Actually I tried useEffect in the first place and because I didn't fully understand how it works I always get the component being re-rendered again and again. Now with your advice and info I can give it another try....

lavocatt:
> Thanks @lavocatt for the advice. Actually I tried useEffect in the first place and because I didn't fully understand how it works I always get the component being re-rendered again and again. Now with your advice and info I can give it another try....

If you need any help along the way let me know, I'd be glad to help!

lavocatt:
@howardgao I've added the commit with the `useEffect` but that's only for the login part, we need also to work on the requests that are preformed on the click of the buttons. But I'm sending this first step so you can have a look tomorrow morning.

gaohoward:
> @howardgao I've added the commit with the `useEffect` but that's only for the login part, we need also to work on the requests that are preformed on the click of the buttons. But I'm sending this first step so you can have a look tomorrow morning.

Thanks a lot @lavocatt !

gaohoward:
> > @howardgao I've added the commit with the `useEffect` but that's only for the login part, we need also to work on the requests that are preformed on the click of the buttons. But I'm sending this first step so you can have a look tomorrow morning.
> 
> Thanks a lot @lavocatt !

Is there some way to remember the login state when I navigate away from the broker details page and then back it does login again.

gaohoward:
> > @howardgao I've added the commit with the `useEffect` but that's only for the login part, we need also to work on the requests that are preformed on the click of the buttons. But I'm sending this first step so you can have a look tomorrow morning.
> 
> Thanks a lot @lavocatt !

Is there some way to remember the login state when I navigate away from the broker details page and then back it does login again.

> noying. But if we remove the alert popup and the login goes silent, it could be ok to login on every re

I don't think login on every refresh is a good idea. 

The login alert is there for now it's going to be replaced by a real login form (when we have a good idea about that). For now it's just a placeholder. And the `check-jolokia` tab is just for dev purpose. In production it should be removed.

lavocatt (approved):
All is good for me. I'd like to work on a follow up for how the requests are sent to the api-server especially in regards of error handling, I'll create a PR in the coming days for that.

PR #127, opened 2024-04-24
Fixed error while running the plugin in the local development env using remote credentials

Msarawan:
on the remote cluster, there is a broker created without `spec` field. it checks the `spec` field as it's not defined in the broker it prompts error with undefined `deployment` and undefined `size ` on the console.
with some adjustments in the code the issue is fixed.

### DIFF
--- a/src/brokers/view-brokers/components/BrokersList/BrokerRow.tsx
+++ b/src/brokers/view-brokers/components/BrokersList/BrokerRow.tsx
@@ -31,11 +31,10 @@ export const BrokerRow: FC<BrokerRowProps> = ({
   const { t } = useTranslation();
   const {
     metadata: { name, creationTimestamp, namespace },
-    spec: {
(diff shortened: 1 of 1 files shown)


src/configuration/Configuration.container.tsx
```diff
@@ -28,6 +29,12 @@ const ConfigurationContainer: FC<ConfigurationContainerProps> = ({
       )
     : null;
 
+  if (!configurationSettings?.spec?.deploymentPlan.image) {
```
lavocatt, line 32: We want to test if the `deploymentPlan` exists, use this instead:
`if (!configurationSettings?.spec?.deploymentPlan) {`

Msarawan: yes, i will try that once i fix the error with plugin running

lavocatt (approved):
Nice finding!

PR #130, opened 2024-04-29
use react-query to perform the network requests

lavocatt:
This work is a working proposal to replace the manual requests to the `api-server` by generated code using the `react-query` library.

**why we want to introduce `react-query`**

There's plenty of blog posts out there venting the merits of delegating to a third party library data fetching. We could sum them up as: it's really hard to produce such code the right way.

Fetching data, caching it, handling fetching errors, retries, reloading it periodically, handling it's lifecycle... It really can lead to difficult debugging sessions and unoptimized UX behaviors.

**new dependencies for the project**

Since OpenShift is running React version 17, we're settling for  `react-query` at version `4`. The documentation is there: 
[React-query@v4](https://tanstack.com/query/v4/docs/framework/react/overview)

React-query also supports code generation. Code generation turns the swagger definition of the api-server into ready to use functions. It's done by the project https://github.com/7nohe/openapi-react-query-codegen.

**what it changes for the development workflow in regards to network calls**

The development workflow now puts the swagger openapi.yml file at the center of the interactions between the `api-server` and the frontend. When a contributor wants to update or add a new endpoints, they now have these actions to perform:

1. update the yaml definition of the api
2. run `yarn build-api-doc`
3. run `yarn codegen`
4. update/code the endpoints on the api-server side
5. use the related hooks on the frontend-side

> [!NOTE] 
> to reviewers:
> The files under `src/openapi` are generated. The core of the contributions is in `src/utils/api-server.tsx` where the `useJolokiaLogin` and  `TestJolokiaPanel` were upgraded.


lavocatt:
Actually the last commit could be in a follow up PR. Let me know if that makes sense to split it or not.

gaohoward:
@lavocatt These is an error in the api server [here](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/blob/c107594f3d16c6f952f7a113c4651cdcaa0b838b/api-server/api/controllers/api_impl.ts#L152). It extracts `params` from request but the actual name is `arguments` in [openapi def](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/blob/c107594f3d16c6f952f7a113c4651cdcaa0b838b/api-server/config/openapi.yml#L702)

Can you fix that? Maybe use some names other than 'arguments` as the word seems give compilation error on my local. Perhaps `args` will do. Just fyi.

api-server/config/openapi.yml
```diff
@@ -113,7 +113,7 @@ paths:
       requestBody:
         required: true
         content:
-          application/x-www-form-urlencoded:
+          application/json:
```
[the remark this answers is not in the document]
lavocatt: I've chose to change it to json because I had a bit of an issue with the parameter serialization between the code generated routine and the api-server. IIRC the request was correctly populated by `x-www-form-urlencoded ` variables, but the server couldn't recover them for some reason. Switching to `json` solved the issue.

It shouldn't pose any issue for the login form, as the values would be passed to the login hook as they are now done.

src/utils/api-server.tsx
```diff
@@ -234,6 +132,7 @@ const JolokiaTestPanel: FC<JolokiaTestPanelType> = ({
         method: 'GET', // *GET, POST, PUT, DELETE, etc.
         headers: {
           'jolokia-session-id': authToken,
+          'jolokia-target-pod-session-id': ordinal.toString(),
```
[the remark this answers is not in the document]
lavocatt: I'll clean all of this then

src/openapi/jolokia/queries/index.ts
```diff
@@ -0,0 +1,489 @@
...
+        jolokiaSessionId: string;
+        jolokiaTargetPodSessionId: number;
+        requestBody: OperationRef;
+      },
+      unknown
+    >,
+    'mutationFn'
+  >,
+) =>
+  useMutation(
```
Msarawan, line 186: @lavocatt , what's the useMutation hook does here?

lavocatt: There is two different kind of hooks provided by `react-query`. The mutations and the queries. Mutations are used to perform an operation that might cause changes on the server, queries are used to gather information from the server. You can see queries as something that needs to be refetched periodically, vs a mutation is something to execute only when needed.

In the PR, a mutation is used for login as it's performing a server side operation and a query is used to get the api-info since this operation is only gathering data from the server to the client.

Maybe some better explanation over there:
https://tanstack.com/query/v4/docs/framework/react/guides/queries
https://tanstack.com/query/v4/docs/framework/react/guides/mutations

api-server/config/openapi.yml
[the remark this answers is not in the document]
lavocatt: See, when we call the mutation to login https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/pull/130/files#diff-43a4379a03fa59a3cab4d52b04b5eaa1df6059c2107e28cc1db0d0e88fed3e53R397 we're passing the necessary values as a parameter.
If these would be gotten from a form it would mean that on `onsubmit` we would call the mutation passing the form values in the `requestBody` object.

lavocatt:
> Can you fix that? Maybe use some names other than `arguments` as the word seems give compilation error on my local. Perhaps `args` will do. Just fyi.

I've had it fixed! Thanks for figuring it out.

api-server/config/openapi.yml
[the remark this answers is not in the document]
gaohoward: ok understand now. Thanks for explanation

src/openapi/jolokia/queries/index.ts
Msarawan: Okay, got it. Thanks for the explanation.

lavocatt:
> > Actually the last commit could be in a follow up PR. Let me know if that makes sense to split it or not.
> 
> I'm fine with either option. :)

After rebasing onto main, I needed to redo the structure of the commits, I've decided, to make it easier on me, to integrate the former last commit into the main contribution.

api-server/api/controllers/security.ts
```diff
@@ -180,8 +180,14 @@ export const VerifyLogin = async (
...
               const brokerKey = decoded['id'];
               const jolokia = securityStore.get(brokerKey);
-              res.locals.jolokia = jolokia;
-              next();
+              if (jolokia) {
+                res.locals.jolokia = jolokia;
+                next();
+              } else {
+                res.status(401).json({
+                  message: 'This session has expired. Please login again',
```
Msarawan, line 188: Hello @lavocatt , LGTM, just small suggestion here. instead of adding direct text here could you please update this with using `useTranslation` hook?
`message: {t('the_session_expired_login_again')}`
or-else any best message you think.

lavocatt: Do we use the translation service for the `api-server` ?

Msarawan: yes you're correct. translation services are not commonly used in api-server. but, there are some scenarios where translation might be necessary in an API context, such as: showing Error Messages, and Response Data if it includes user-visible text.

lavocatt: That makes sense. I'll make a patch to translate all the messages coming from the api-server.

gaohoward (dismissed):

gaohoward:
@lavocatt Hi Thomas, when I pulled your branch and run `yarn codegen` it generated some changes. Can you verify it is the case for you too?

Msarawan:
> @lavocatt Hi Thomas, when I pulled your branch and run `yarn codegen` it generated some changes. Can you verify it is the case for you too?

`+1` same for me as well.

Msarawan (approved):
Good work! @lavocatt

lavocatt:
> > @lavocatt Hi Thomas, when I pulled your branch and run `yarn codegen` it generated some changes. Can you verify it is the case for you too?
> 
> `+1` same for me as well.

Yes, that's an issue. If you run `yarn codegen` `npx pretty-quick` and `yarn lint` there should be only two changes related to imports I manually removed because they are causing compilation errors.

I will add an eslint plugin to remove the unused imports automatically and then update the codegen command to use the 3 commands in a row.

lavocatt:
I've added:
* an eslint plugin to automatically clean the unused imports
* the prettier and the linter to the codegen command
* a check in the ci to make sure the generated files are in sync with the `openapi.yml` file

lavocatt:
After some discussion with @Msarawan we've decided to drop the commit about internationalizing the api-server for now and to discuss the topic in the https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/discussions

Msarawan:
> I've added:
> 
> * an eslint plugin to automatically clean the unused imports
> * the prettier and the linter to the codegen command
> * a check in the ci to make sure the generated files are in sync with the `openapi.yml` file

That's nice!, now `yarn codegen` works fine without any warnings or errors.

gaohoward:
All good. Nice job!

gaohoward (approved):

PR #139, opened 2024-05-13
Enable Express hot reloading

Msarawan:


lavocatt:
Hot reloading is now working natively with webpack, we can close this PR. Thanks anyway for the time you spent on this!

PR #140, opened 2024-05-14
[Addresses page]: Integrate API for address data

Msarawan:
Integrate the design and implementation of the address page into the top-level entry on the Broker details page.

Screenshot from 2024-05-22 18-46-09


src/addresses/Address.component.tsx
```diff
@@ -0,0 +1,58 @@
+import { useTranslation } from '../i18n';
+import {
+  TableColumn,
+  VirtualizedTable,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { Address } from './Address.container';
```
lavocatt, line 6: The `Address` type will be imported from the generated code.

src/addresses/Address.container.tsx
```diff
@@ -0,0 +1,44 @@
+import { FC, useContext, useEffect, useState } from 'react';
+import { Addresses } from './Address.component';
+import { AuthContext, K8sResourceCommon } from '../utils';
+import { useJolokiaServiceGetAddresses } from '../openapi/jolokia/queries';
+
+export type Address = {
+  name: string;
+};
```
lavocatt, line 8: You can remove this as we will rely on the code generation to obtain this type.

api-server/config/openapi.yml
```diff
@@ -810,3 +810,12 @@ components:
...
         additionalProperties: true
+    BrokersAddressResponse:
+      type: array
+      items:
+        type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
```
lavocatt, line 821: Let's change this to:
```
    Address:
      type: object
      required:
        - name
      properties:
        name:
          type: string
```

src/addresses/AddressRow.tsx
```diff
@@ -0,0 +1,27 @@
+import { FC } from 'react';
+import {
+  RowProps,
+  TableData,
+  TableColumn,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { Address } from './Address.container';
```
lavocatt, line 7: The Address will be imported from the codegen

src/brokers/broker-details/BrokerDetails.container.tsx
```diff
@@ -20,8 +20,9 @@ import { useTranslation } from '../../i18n';
 import {
   ClientsContainer,
   ConfigurationContainer,
-  QueuesContainer,
-  TopicsContainer,
+  AddressContainer,
+  // QueuesContainer,
+  // TopicsContainer,
```
lavocatt, line 25: It's better to not leave commented lines

src/brokers/broker-details/BrokerDetails.container.tsx
```diff
@@ -142,12 +143,18 @@ const BrokerDetailsPage: FC = () => {
...
+          >
+            <AddressContainer />
+          </Tab>
+          {/* <Tab eventKey={3} title={<TabTitleText>{t('queues')}</TabTitleText>}>
             <QueuesContainer />
           </Tab>
           <Tab eventKey={4} title={<TabTitleText>{t('topics')}</TabTitleText>}>
             <TopicsContainer />
-          </Tab>
+          </Tab> */}
```
lavocatt, line 157: same here for the commented lines.

lavocatt (commented):
You are making good progress :rocket: 

During our pair programming session we've introduced an important change in the `openapi.yaml` definition. However the way we wrote it was slightly suboptimal.

See we've declared a `BrokersAddressResponse` type that was an Array of properties and used that as the return value of the `addresses` endpoint. As a result the codegen produced a very generic Typescript type:
`export type BrokersAddressResponse = Array<Record<string, any>>;`

Instead we should have declared an `Address` type and say that the `addresses` endpoint would return an array of `Address` objects. This will make the codegen create a much more useful type for us:

```
export type Address = {
  name: string;
};
```

This looks exactly like the type that you created manually in the `Address.container.tsx` file!

At the moment, we still have full control over the `api-server` and we can make it return any type that is convenient to us, so let's leverage that.

Something to keep in mind is that we shouldn't use useEffects when we do things that are purely within React. I've added a pointer to the documentation, it's worth reading.

Msarawan:
> You are making good progress 🚀
> 
> During our pair programming session we've introduced an important change in the `openapi.yaml` definition. However the way we wrote it was slightly suboptimal.
> 
> See we've declared a `BrokersAddressResponse` type that was an Array of properties and used that as the return value of the `addresses` endpoint. As a result the codegen produced a very generic Typescript type: `export type BrokersAddressResponse = Array<Record<string, any>>;`
> 
> Instead we should have declared an `Address` type and say that the `addresses` endpoint would return an array of `Address` objects. This will make the codegen create a much more useful type for us:
> 
> ```
> export type Address = {
>   name: string;
> };
> ```
> 
> This looks exactly like the type that you created manually in the `Address.container.tsx` file!
> 
> At the moment, we still have full control over the `api-server` and we can make it return any type that is convenient to us, so let's leverage that.
> 
> Something to keep in mind is that we shouldn't use useEffects when we do things that are purely within React. I've added a pointer to the documentation, it's worth reading.

sure, Thanks @lavocatt !

api-server/api/controllers/api_impl.ts
```diff
@@ -78,10 +78,17 @@ export const getAddresses = (
...
       .then((result: any[]) => {
-        res.json(result);
+        const names = result.map((r) =>
+          r.split(',')[0].split('=')[1].replace(new RegExp('"', 'g'), ''),
+        );
+        const returnValue = names.map((name) => {
+          return { name: name };
+        });
+        res.json(returnValue);
+        console.log('returnValue', returnValue);
```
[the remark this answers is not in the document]
Msarawan: sure

api-server/config/openapi.yml
```diff
@@ -456,7 +456,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/BrokersResponse'
+                $ref: '#/components/schemas/BrokersAddressResponse'
```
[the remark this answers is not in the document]
Msarawan: yeah, make sense to me. thanks!

src/brokers/broker-details/BrokerDetails.container.tsx
Msarawan: yes, i will remove and update the PR.
Thanks!

src/addresses/Address.container.tsx
```diff
@@ -1,33 +1,26 @@
...
 
   useEffect(() => {
-    if (isSuccess) {
-      const AddressData = rawAddresses.map((address) => {
-        return {
-          name: address.name,
-        };
-      });
+    if (isSuccess && rawAddresses) {
+      const AddressData = getAddressData(rawAddresses);
```
lavocatt, line 23: One last thing to fix, now that you've updated the yaml, there's no need to transform `rawAddresses` to something else. Just send it to the `addressData` parameter of the `Addresses` component.

Also it's an opportunity to rename `rawAddresses` to `addrresses` since it's not raw data that needs to be parsed anymore.

Msarawan: okay sure, i will update.

gaohoward:
The page looking good to me. I think we may add some links to each address to let user navigate to address details or some action items to make this page more useful. But I think that subjects to further discussions before doing that, and in more PRs.

gaohoward (approved):

Msarawan:
> The page looking good to me. I think we may add some links to each address to let user navigate to address details or some action items to make this page more useful. But I think that subjects to further discussions before doing that, and in more PRs.

Thanks @gaohoward . Yes, that part i continue with my upcoming PR. where onclick on each address we move to the specific address details page and we can show the users some useful details about each address.

lavocatt (approved):

Issue #141, opened 2024-05-16
Adding new api /readAddressAttributes


gaohoward:
Add an api to read Address attributes

PR #142, opened 2024-05-16
[#141] Adding new api /readAddressAttributes

gaohoward:


api-server/config/openapi.yml
```diff
@@ -298,6 +298,69 @@ paths:
...
+          required: true
+          in: query
+        - name: attrs
+          description: attribute names separated by commas. If not speified read all attributes.
+          required: false
+          in: query
+          schema:
+            type: array
+            items:
+              type: string
```
lavocatt, line 354: Can you provide the list of valid attributes in the specification as an enumeration please?
This will help:
* to know in the frontend what are the valid options to request
* to limit the scope of the endpoint to values we have control on

It should look something like this:
```
attrs:
  type: array
  description: attribute names separated by commas. If not specified read all attributes.
  items:
    type: string
    enum:
      - RoutingTypes
```

api-server/config/openapi.yml
```diff
@@ -298,6 +298,69 @@ paths:
...
+          style: form
+          explode: false
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BrokersResponse'
```
lavocatt, line 363: Having a more detailed schema will make the code generation do the Typescript work for us:

```
      responses:
        200:
          description: Success
          content:
            application/json:
              schema:             
                type: array
                items:
                  $ref: '#/components/schemas/AddressAttribute'
```

with at the end of the document a new `AddressAttribute` definition:

```
    AddressAttribute:
      type: object
      properties:
        request:
          type: object
          properties:
            mbean:
              type: string
            attribute:
              type: string
            type:
              type: string
        value:
          type: array
          items:
            type: string
        timestamp:
          type: number
        status:
          type: number
```

api-server/config/openapi.yml
gaohoward: @lavocatt Hi, this information can be dynamically discovered by a query to the api server, e.g.
`https://localhost:9443/api/v1/addressDetails?name=DLQ`
It will return all the operations and attributes available on address `DLQ`.

api-server/config/openapi.yml
gaohoward: That's nice to have. Currently `BrokerResponse` is used as a common response schema for almost every api. I think we can do this for all other apis not only for this api. I've create a issue for that:
https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/146

api-server/config/openapi.yml
```diff
@@ -298,6 +298,69 @@ paths:
...
+          ]
+          ```
+          **Note**: to read multiple attributes, set it to **attrs** parameter
+          separated by commas, e.g. `RoutingTypes,Address`.
+      tags:
+        - jolokia
+      operationId: readAddressAttributes
+      parameters:
+        - in: header
+          name: jolokia-session-id
```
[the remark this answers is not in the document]
gaohoward: +1. Will do. Thanks!

Msarawan (approved):
Looks Good!

api-server/config/openapi.yml
lavocatt: The question would be, how often are we expecting this list to change?

Is it truly dynamic in the sense that it is defined by the runtime, or is it somehow a static content that is served every time by the jolokia endpoint?

If I remember correctly, the last time I called `addressDetails` on a couple of addresses, the returned object was always the same. My call would be that if we can expect this object to always be the same, then we should treat it as API definition.

There are multiple benefits of having a limited set of possibilities, the main of which is to keep the scope under control. The second by extension is to limit any abusive use by the user.

api-server/config/openapi.yml
lavocatt: This PR is adding a new endpoint, it makes sense to have it already returning a tailored result, #140 would greatly benefit of this tailoring.

I agree that a followup PR is necessary for all the other endpoints in any case.

lavocatt (approved):
All looks good to me. Thanks for you work on this.
I will work on a follow up PR to address #146 and to include some of suggestions this afternoon.

api-server/config/openapi.yml
gaohoward: For each compnent the operations and attributes are defined in the broker, for example here is the definition for address component https://github.com/apache/activemq-artemis/blob/2.33.0/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
(it's documented in the api.md https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/blob/main/api.md#getaddressdetails)

The definition may change over time but mostly the change is additive, i.e. new ops or attrs may be added in future releases but existing ones won't normally be changed.

Access control is a rather larger topic and not fit into this PR. First the console itself has a login, our plugin may adopt that, or have our own authorization framework. Second the jolokia access can be configured on the broker side. Here: https://github.com/apache/activemq-artemis/blob/2.33.0/docs/user-manual/management.adoc#configuring-jmx

api-server/config/openapi.yml
gaohoward: https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/146 is a common issue, and it should be solved separately. I would suggest not to block this PR by it. The major purpose of this PR is add a new endpoint. Not to partially solve another issue.

gaohoward:
@lavocatt Thanks Thomas!

PR #149, opened 2024-05-28
Refactor OpenAPI Documentation: Remove Redundant Examples from Descriptions

Msarawan:
Updated openapi.yml file.


lavocatt:
I'm wondering if we shouldn't instead remove this information from the description of the endpoints.
We can rely on the actual definition to know what is the url.
This is also true for the example input & output. Since they are typed.

lavocatt:
I'm usually quite up for having documentation as much as possible. And if possible I'd like the documentation not to be just paraphrasing the code, it should add something more. Because if it don't, it kinda means that we are writing the code twice.

My take is that we should first merge #148 then have a pass on the openapi doc to make sure it is sound.

lavocatt:
I think this looks better now! Thanks!
You'll need to regenerate the `api.md` file though `yarn build-api-doc`.

lavocatt:
> I think this looks better now! Thanks! You'll need to regenerate the `api.md` file though `yarn build-api-doc`.

Oups, it was already there! I'm saying nonsense :smile:

lavocatt (approved):
All looks good! Thanks for making our documentation consistent with the code. That's a good initiative!

PR #156, opened 2024-06-03
Broker view with reducer

lavocatt:
@howardgao here's the rework with the useReducer


lavocatt:
Worked rebased, closing the draft PR.

PR #157, opened 2024-06-05
[#163] Address Details page

Msarawan:
Address Details page:
Screenshot from 2024-06-13 16-05-00

### DIFF
--- a/src/addresses/address-details/AddressDetails.container.tsx
+++ b/src/addresses/address-details/AddressDetails.container.tsx
@@ -1,0 +1,135 @@
+import { FC, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { AddressDetailsBreadcrumb } from '../../shared-components/AddressDetailsBreadcrumb/AddressDetailsBreadcrumb';
+import {
+  PageSection,
+  PageSectionVariants,
+  Title,
+} from '@patternfly/react-core';
+import { useTranslation } from '../../i18n';
+import {
+  AMQBrokerModel,
+  AuthContext,
+  getApiServerBaseUrl,
+  useJolokiaLogin,
+} from '../../utils';
+import {
+  K8sResourceKind,
+  k8sGet,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { OpenAPI as OpenAPIConfig } from '../../openapi/jolokia/requests/core/OpenAPI';
+import { AddressDetails } from './AddressDetails.component';
+
+const AddressDetailsPage: FC = () => {
+  const { t } = useTranslation();
+  const {
+    name,
+    ns: namespace,
+    brokerName,
+    podName,
+  } = useParams<{
+    name?: string;
+    ns?: string;
+    brokerName?: string;
+    podName?: string;
+  }>();
+  const [brokerDetails, setBrokerDetails] = useState<K8sResourceKind>({});
+  const [_loading, setLoading] = useState<boolean>(true);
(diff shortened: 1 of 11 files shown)


lavocatt (commented):
That's good work!

I've got some feedback on the breadcrump. As you can see on the image below, the user can only go back to Addresses, but I would like the user to go back to other leves above as well.

image

The full path should be in the breadcrumb:

[Brokers](http://localhost:9000/k8s/ns/default/brokers) > [default](http://localhost:9000/k8s/ns/default/brokers/default) > [default-ss-0](http://localhost:9000/k8s/ns/default/broker-pods/default/default-ss-0) > [Addresses](http://localhost:9000/k8s/ns/default/broker-pods/default/default-ss-0?tab=Addresses) > $AddressName

* Clicking on `brokers` takes to the list of brokers
* Clicking on `default` takes to the list of pods for the broker
* Clicking on the `default -ss-0` takes to the brokers details of this pod
* Clicking on `Addresses` takes to the addresses tab of the broker details
* Then the name of the address gets displayed at the end

src/addresses/address-details/AddressDetails.container.tsx
```diff
@@ -0,0 +1,133 @@
...
+        console.error(e);
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  };
+
+  useEffect(() => {
+    k8sGetBroker();
+  }, []);
```
lavocatt, line 66: To execute something on first mount, instead of using `useEffect` use this pattern:

```typescript
const [isFirstMount, setIsFirstMount] = useState(true);
if (isFirstMount) {
  k8sGetBroker();
  setIsFirstMount(false);
}
```

src/addresses/address-details/AddressDetails.container.tsx
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  const [prevIsLoading, setPrevIsLoading] = useState(isLoading);
+  const [notify, setNotify] = useState(false);
+  if (prevIsLoading !== isLoading) {
+    if (!isLoading && source === 'api') {
+      setNotify(true);
+    }
+    setPrevIsLoading(isLoading);
+  }
```
lavocatt, line 83: Good usage of a previous value to avoid using `useEffect` here :partying_face:

src/addresses/address-details/AddressDetails.container.tsx
```diff
@@ -0,0 +1,133 @@
...
+      // TODO maybe use the OpenShift console notification system to let the
+      // user know that the login was a success?
+      alert(
+        'login successful ' + brokerDetails?.metadata?.name + ' token ' + token,
+      );
+    } else {
+      alert('login failed');
+    }
+    setNotify(false);
+  }
```
lavocatt, line 95: We need to figure a way to notify the user without alerts, but that's out of the scope of this PR.

Msarawan: yes.

gaohoward: There would be a login solution when we have a good idea how to to authentication/authorization.

gaohoward:
The breadcrumb link doesn't update correctly after I clicked an address.
Before it is like
before
After entering address page:
after

gaohoward:
A suggestion to add a 'search field` to the attributes page as it has quite a few items across multiple pages:
search-box-suggestion

lavocatt:
That's good work! I think @gaohoward's suggestion about the search field is a very good idea.

lavocatt (dismissed):
That's a very good piece of work! Congratulations.

Msarawan:
> That's a very good piece of work! Congratulations.

Thanks! @lavocatt

gaohoward (dismissed):

Issue #147, opened 2024-05-20
Implementing Broker Creation form


gaohoward:
Currently we provide a `form view` and 'yaml view` when creating a broker via the plugin.

The `From view` need to be implemented to help user
* doing configuration via a visual page rather than editing raw yamls
* focus on broker configuration components like acceptors without worrying use deprecated CR fields
* assist user on SSL/TLS setup

Leaving yaml editing for advanced users (like use resourceTemplate to path the resources)

PR #161, opened 2024-06-07
[#147] Adding show cert details panel

gaohoward:


src/configuration/CertificateDetailsModal.tsx
```diff
@@ -0,0 +1,358 @@
...
+        result += crt.issuer.attributes[i].shortName;
+        result += ' = ';
+        result += crt.issuer.attributes[i].value;
+        if (i !== crt.issuer.attributes.length - 1) {
+          result += ', ';
+        }
+      }
+    }
+    return result;
+  };
```
lavocatt, line 41: In JS we tend to use the foreach syntax when available:

```typescript
  const formatCertIssuer = (crt: pki.Certificate) => {
    let result = '';
    crt.issuer.attributes.forEach((attribute, i, all) => {
      result += attribute.shortName;
      result += ' = ';
      result += attribute.value;
      if (i !== all.length - 1) {
        result += ', ';
      }
    });
    return result;
  };
```

One main advantage is that you don't have to "dig" in for the `i`th attribute in the array yourself.

src/configuration/CertificateDetailsModal.tsx
```diff
@@ -0,0 +1,358 @@
...
+        result += crt.subject.attributes[i].shortName;
+        result += ' = ';
+        result += crt.subject.attributes[i].value;
+        if (i !== crt.subject.attributes.length - 1) {
+          result += ', ';
+        }
+      }
+    }
+    return result;
+  };
```
lavocatt, line 56: Same comment about the for loop to foreach.

src/configuration/CertificateDetailsModal.tsx
```diff
@@ -0,0 +1,358 @@
...
+                      <code>{getExtPropContent(ext, key)}</code>
+                    </SplitItem>
+                  </Split>
+                ),
+            ),
+          ]}
+        </>
+      )),
+    ];
+  };
```
lavocatt, line 100: This function must be a React component as it is producing JSX.

```typescript
interface ExtensionsListProps {
  crt: pki.Certificate;
  order: number;
}

const ExtensionsList: FC<ExtensionsListProps> = ({ crt, order }) => {
  //https://datatracker.ietf.org/doc/html/rfc5280
  const getExtPropContent = (ext: any, key: string) => {
    let result = '';
    if (typeof ext[key] === 'object') {
      result += JSON.stringify(ext[key]);
    } else {
      result += ext[key].toString();
    }

    return result;
  };
  return (
    <>
      {crt.extensions.map((ext, index) => (
        <>
          <Split key={'extension' + index + 'order' + order}>
            <SplitItem>{index + 1} -</SplitItem>
            <SplitItem isFilled>
              <b>{ext.name}</b>
            </SplitItem>
          </Split>
          {[
            Object.keys(ext).map(
              (key) =>
                key !== 'name' &&
                key !== 'value' && (
                  <Split hasGutter key={ext.name + key}>
                    <SplitItem></SplitItem>
                    <SplitItem>
                      <b>{key}: </b>
                    </SplitItem>
                    <SplitItem isFilled>
                      <code>{getExtPropContent(ext, key)}</code>
                    </SplitItem>
                  </Split>
                ),
            ),
          ]}
        </>
      ))}
    </>
  );
};
```

src/configuration/CertificateDetailsModal.tsx
```diff
@@ -0,0 +1,358 @@
...
+      }
+    }
+
+    return result;
+  };
+
+  const formatCertInfo = (crt: pki.Certificate, order: number) => {
+    return (
+      <>
+        <Split hasGutter key={'version' + order}>
```
lavocatt, line 170: `Split` is a layout organizer. Instead it would be better to use components, [`DescriptionList`](
 http://v4-archive.patternfly.org/v4/components/description-list) would be a good candidate.
 
 Why choose components instead of layouts? I think the reason is that PatternFly takes care of the design for us. If we pick components, we ensure that we're having a coherent visual with every RedHat product out there.

While if we're using layouts, we're on our own regarding style, and that would mean in the end to spend more time & energy to fit the look and feel of the portfolio.

src/configuration/CertificateDetailsModal.tsx
```diff
@@ -0,0 +1,358 @@
...
+  const getCertDetails = () => {
+    return (
+      <>
+        {certs.length > 0 &&
+          certs.map((cert, index) => (
+            <StackItem key={index}>{formatCertInfo(cert, index)}</StackItem>
+          ))}
+      </>
+    );
+  };
```
lavocatt, line 323: Same as for `getExtensionsList` this must be a react component since it is returning JSX:

```typescript
  const getCertDetails = () => {
    return (
      <>
        {certs.length > 0 &&
          certs.map((cert, index) => (
            <StackItem key={index}>{formatCertInfo(cert, index)}</StackItem>
          ))}
      </>
    );
  };
```

lavocatt (commented):
Good work, this is making good progress. I have comments on the technicalities, don't hesitate to reach out to me if you need some help.

src/configuration/CertificateDetailsModal.tsx
lavocatt: It's overall better to extract as much as named components as possible instead of having a list too many "indentation" levels of components.

Here's a side good effect from this coding style:

image

In the hierarchy of components we find our Extension and ExtensionList and ExtensionKey sub components with the parameters they got. Quite handy when debugging is needed :slightly_smiling_face:

gaohoward:
@lavocatt Hi I've updated the branch. Please take a look. Thanks!

lavocatt:
I really like how this is looking now! Good work @howardgao :rocket: 
image

Issue #150, opened 2024-05-28
Implement a PoC for tls enabled coms with cert manager generated certs


lavocatt:
Implement: https://github.com/artemiscloud/activemq-artemis-operator/pull/927 on the provisioning plugin so that we can hook up a couple of external producers & consumers to the broker instance.

PR #162, opened 2024-06-07
[#150] cert-manager PEM's with ingress annotation 

lavocatt:
Adds entries in the broker configuration form to create a cert-manager annotation on an acceptor.

Steps to use it:

## Create an acceptor
image

## Create an annotation
image

## Add a chain of trust if necessary
image

## Annotation is created
image

## Create the broker and go to its details page
image

## Download the certificate and connect to the broker:
image


lavocatt:
This is how it looks now:

image

The user has de possibility to select an issuer when they want to enable SSL. However we've had a discussion on the subject with @howardgao the other day about how this tutorial is integrated. And I do agree with you now Howard that this is not the place for these kind of configuration.

What I propose to do instead is to add another entry under `console` for the `resourceTemplates` and add all the manual knobs for a user to manually set this. And at the same time, to propose a third view in the broker creation.

In this view the user would set choose between a list of supported tutorials. Selecting a tutorial would take the user through the necessary steps to achieve what they want. In this particular case, the first tutorial would be to setup a PEM certificate using cert-manager.

wdyt?

lavocatt:
A follow up PR refactors the UI a bit to use patternfly components: #174

gaohoward:
How about this? This can reduce the space occupied by the `Card` and provide an easy access to the annotation feature for the users.

adding-annotation-entry-in-select

lavocatt:
> How about this? This can reduce the space occupied by the `Card` and provide an easy access to the annotation feature for the users.
> 
> adding-annotation-entry-in-select

Thanks for the feedback, it's true that the annotation card is taking some space. I'm not sure if "hiding" it under the certs is really semantically equivalent. The annotation itself does more than just generating certs at runtime, it also expose the acceptor through an ingress.

I'm going to work on something slightly different, instead of showing all the available cards under annotation, I'll add a button in the general configuration to add an annotation, clicking on the button will bring a modal where the user can select which annotation to get.

Then if an annotation gets selected, it'll be displayed in the list under.

lavocatt:
Now the user can click on a button to add an annotation 
image

It opens up a modal where it can select the kind of annotation they want.
image

When the annotation is added it is listed as part of the configuration:
image

RoddieKieley:
@lavocatt I like the simplification of the interface that is immediately displayed to the user, however as a messaging user I may not automatically think 'automatic certificate generation' when I see an Add Annotations button. The annotations are an implementation detail whereas what I'd be looking for once I toggled SSLEnabled to be turned on is an option to automatically generate the certificate.

Maybe the user needs a hint of some sort to point them in the direction of the button, or maybe it needs to use different verbiage that more clearly indicates that you can further configure the SSL/TLS for automatic certificate generation and usage of the cert-manager operator.

gaohoward (approved):

PR #165, opened 2024-06-10
Add BrokerDetails breadcrumb page

Msarawan:
Added BrokerDetails breadcrumb page and updated BrokerPods breadcrumb page.

Screenshot from 2024-06-11 01-47-31

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -42,8 +42,8 @@
     "properties": {
       "exact": false,
       "path": [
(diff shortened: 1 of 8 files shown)


Msarawan:
> Good work! A nice follow up would be to find a way to generalize things a bit. Like for instance, the breadcrumb doesn't only contain the breadcrump but also buttons to edit the broker, if someone wants to update the buttons now, they have to do it in two different places. Not an immediate issue, but something to think about soon.

Thanks! 
yeah make sense.

PR #166, opened 2024-06-11
Share editor views

gaohoward:


src/brokers/utils/add-broker.ts
```diff
@@ -168,7 +212,9 @@ type ArtemisReducerActions =
   | SetNamespaceAction
   | SetReplicasNumberAction
   | UpdateAcceptorFactoryClassAction
-  | UpdateConnectorFactoryClassAction;
+  | UpdateConnectorFactoryClassAction
+  | UpdateEditorStateAction
+  | SetModelAction;
```
lavocatt, line 217: The values are alphabetically ordered, can you place this new `SetModelAction` before `SetReplicasNumberAction` please?

src/brokers/utils/add-broker.ts
```diff
@@ -129,6 +169,10 @@ export enum ArtemisReducerOperations {
   updateAcceptorFactoryClass,
   /** Updates the configuration's factory Class */
   updateConnectorFactoryClass,
+  /** updates the current state item */
+  updateEditorState,
+  /** updates the whole model */
+  setModel,
```
lavocatt, line 175: The enumeration is alphabetically ordered, can you place this `setModel` before ` setReplicasNumber` please?

src/brokers/add-broker/components/AddBrokerForm/YamlEditorView.tsx
```diff
@@ -48,19 +45,15 @@ const YamlEditorView: FC<YamlEditorViewProps> = ({
...
   //event: contains information of changes
   //value: contains full yaml model
   const onChanges = (newValue: any, _event: any) => {
-    //yamlValue.yamlData = JSON.parse(newValue);
-    //instead of blatantly replce the whole contents
-    //compare them and only accept additive contents, or
-    // else warning and refuse to update.
-    console.log('old value:', fromState.cr);
-    console.log('new value', JSON.parse(newValue));
+    //use reducer?
```
lavocatt, line 55: I think yes, it would be good to use the new `setModel` operation you've introduced.

src/brokers/add-broker/components/AddBrokerForm/YamlEditorView.tsx
```diff
@@ -85,11 +78,14 @@ const YamlEditorView: FC<YamlEditorViewProps> = ({
             <CodeEditor
               value={JSON.stringify(fromState.cr, null, '  ')}
```
lavocatt, line 79: Could be worth it to transform the `json` to yaml here. That way the user would have a consistent feel with other yaml editors.

src/brokers/utils/add-broker.ts
```diff
@@ -18,8 +22,40 @@ export const BrokerCreationFormState = createContext<FormState>({});
...
+  const key = name + ns;
+  let formState = artemisCRStateMap.get(key);
+  if (!formState) {
+    formState = {};
+    formState.shouldShowYAMLMessage = true;
+    formState.editorType = EditorType.BROKER;
+    formState.editorState = {
+      stringValues: new Map<string, string>(),
+      booleanValues: new Map<string, boolean>(),
+    };
```
lavocatt, line 49: I understand that this is used so that the editor remembers what property is being worked on, so that when the user goes back and forth between the configurator and the yaml editor, the last open view stays open. I think this is a great and important addition.

However, I'm not sure if we should work with open fields in the reducer. As it means that we're extracting the logic out of the editor. See Later on in the code the state is changed as followed:

```typescript
  const onSelectBrokerConfigItem = (
    selectedItem: any,
    selectedItemProps: any,
  ) => {
    console.log('new selection selected', selectedItem);
    dispatch({
      operation: ArtemisReducerOperations.updateEditorState,
      payload: {
        stateType: StateType.String,
        itemId: configId,
        state: selectedItemProps.itemId,
      },
    });
  };
```
The operation `updateEditorState` being generic, we can't know what is being changed on the state unless performing a complete examination of the code.

Instead I would prefer to have:

```typescript
  const onSelectBrokerConfigItem = (
    selectedItem: any,
    selectedItemProps: any,
  ) => {
    dispatch({
      operation: ArtemisReducerOperations.setEditorActiveProperty,
      payload: selectedItemProps.itemId,
      },
    });
  };
```

In this case, we the operation is named according to the action it does, and the payload can be documented accordingly to: "set the active property the user is working on."

lavocatt (commented):
I think this is a good piece of work and that it does work very well.

However I feel some restrain about adding open field values in the reducer. See my comments below, my approach designing the reducer was to bring a full set of precisely named operations so that when a contributor reads the code they immediately know what is the operation performing on the state.

With open values we're extracting the logic away from the reducer and back in the components where it could endup spread across multiple components, which might lead to more maintenance burden.

src/brokers/utils/add-broker.ts
gaohoward: I think you mean 'before setNamespaceAction` as M is before N. :)

src/brokers/add-broker/components/AddBrokerForm/YamlEditorView.tsx
```diff
@@ -48,19 +51,17 @@ const YamlEditorView: FC<YamlEditorViewProps> = ({
...
-    // else warning and refuse to update.
-    console.log('old value:', fromState.cr);
-    console.log('new value', JSON.parse(newValue));
-    Object.assign(fromState.cr, JSON.parse(newValue));
-    //_.merge(yamlValue.yamlData, newValue);
-
-    console.log('module after merge:', fromState.cr);
+    dispatch({
+      operation: ArtemisReducerOperations.setModel,
+      payload: Object.assign(fromState.cr, JSON.parse(newValue)),
```
lavocatt, line 63: I think just sending `JSON.parse(newValue)` in the payload should be sufficient.

src/brokers/utils/add-broker.ts
```diff
@@ -43,7 +77,10 @@ export const newArtemisCRState = (namespace: string): FormState => {
   return {
     shouldShowYAMLMessage: true,
     editorType: EditorType.BROKER,
-    cr: initialCr,
+    cr: initCr,
+    editorActiveProperties: {
+      activeProperties: new Map<string, any>(),
```
lavocatt, line 82: We can make the type a notch more precise:
```
 activeProperties:{
  itemId: 'acceptors' | 'connectors' | 'console'`;
  expanded: boolean:
}
```

lavocatt (commented):
Thanks for all the changes you've done, I have only two comments left.

src/brokers/utils/add-broker.ts
gaohoward: should `itemId` be unique for each acceptor/connector as there may be multiple acceptor/connectors
If user expands an acceptor the editor need to know which acceptor/connector is expanded

gaohoward: for example I need to remember that **acceptor0** and **acceptor2** are expanded, but **acceptor1** is not
Screenshot from 2024-06-12 23-18-14

lavocatt: I understand now, that makes sense! I thought it was only for the bigger categories. Let's chat about it in today's UX meeting.

lavocatt (approved):
Thanks for splitting the two PRs!

gaohoward:
@lavocatt Thanks! I'll send a following PR for the ui state management.

PR #169, opened 2024-06-13
UI state management

gaohoward:
- To have a way to keep the UI state of the editor so that when users switch views the UI keeps its state.


lavocatt:
@howardgao what should we do with this PR? Do you want to continue working on it or do you prefer to discard it and open an issue for later instead?

PR #170, opened 2024-06-14
Improve design guidelines with useTranslation and PatternFly icons

Msarawan:
Replaced normal text with useTranslation for localization. and Added icons from PatternFly for better UI consistency.


lavocatt:
Can you separate the commit that updates the dependency to the commit that is updating the translations please?

gaohoward (approved):
Nice Job!

Issue #171, opened 2024-06-17
Cert details page cannot parse certs with ECDSA algorithm


gaohoward:
Currently the plugin uses node-forge https://www.npmjs.com/package/node-forge to parse certificates. 
The node-forge only support `RSA` algorithm so it'll be broken if it parses a cert with ECDSA format.
Using @pecular/x509 https://www.npmjs.com/package/@peculiar/x509 seems to be a nice replacement as it supports ECDSA as well as RSA, and it 
has API to assist paring and converting cert details into human readable strings.

PR #172, opened 2024-06-17
[#171] Cert details page cannot parse certs with ECDSA algorithm

gaohoward:
Use @peculiar/x509 to parse certificates


lavocatt:
There's a small issue in the codeBlock used, they are not adapting to the dark screen mode of openshift 
image

src/configuration/CertificateDetailsModal.tsx
```diff
@@ -10,339 +8,307 @@ import {
...
+  return (
+    <DescriptionListGroup>
+      <DescriptionListTerm>
+        <b>
+          {index + 1} - {extTitle}
+        </b>
+      </DescriptionListTerm>
+      <DescriptionListDescription>
+        <CodeBlockCode
+          style={{ backgroundColor: 'white' }}
```
lavocatt, line 74: It's better to not manually style stuff in most cases. Because it'll have side effects elsewhere. For instance, setting the background white is what makes the code blocks unable to show the data appropriately in dark mode.

gaohoward: ok, will remove the style. Thanks!

gaohoward: @lavocatt I updated the PR (removed style and also improved the public key info display)

gaohoward:
image

lavocatt (approved):
This is a very good addition :rocket:

PR #174, opened 2024-06-17
Use patternfly components for the broker form

lavocatt:
We can leverage the patternFly Form component at the root of the
    acceptor and connector configuration page. The Form component handles
    the layout for us and removes the need of flex organizers. Also it's
    easier to have fields aligning properly without having to deal with
    sideeffects when a subcomponent tries to get more space than it was
    doing before.

depends on: #162

### DIFF
--- a/src/configuration/broker-models.tsx
+++ b/src/configuration/broker-models.tsx
@@ -1,9 +1,7 @@
 import {
-  CertIssuerModel,
-  CertModel,
-  K8sResourceKind,
-  SecretModel,
-} from '../utils';
+  k8sCreate,
+  useK8sWatchResource,
+} from '@openshift-console/dynamic-plugin-sdk';
 import {
   Button,
   Divider,
@@ -34,6 +32,9 @@ import {
   Tooltip,
   ValidatedOptions,
 } from '@patternfly/react-core';
+import { SelectOptionObject } from '@patternfly/react-core/dist/js';
+import * as x509 from '@peculiar/x509';
+import base64 from 'base-64';
 import {
   FC,
   PropsWithChildren,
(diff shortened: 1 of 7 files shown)


src/configuration/console-config.tsx
```diff
@@ -96,87 +95,81 @@ export const ConsoleConfigPage: FC<ConsoleConfigProps> = ({ brokerId }) => {
...
-          <FlexItem>
-            <FormGroup
-              label="ExposeMode"
-              fieldId={'console-config-exposemode-formgroup'}
-              key={'formgroup-console-exposemode'}
+    <Form isHorizontal isWidthLimited key={'form' + brokerId}>
+      <Grid hasGutter md={6}>
+        <FormFieldGroup>
+          <FormGroup
+            label="Expose"
```
Msarawan, line 102: Hey Thomas...all changes looks good!
just a small suggestion here can we use useTranslation() hook to change the `label` text aswell?

src/configuration/console-config.tsx
```diff
@@ -96,87 +95,81 @@ export const ConsoleConfigPage: FC<ConsoleConfigProps> = ({ brokerId }) => {
...
+            fieldId={'console-config-expose-formgroup'}
+            key={'formgroup-console-expose'}
+          >
+            <Checkbox
+              label="Expose Console"
+              isChecked={exposeConsole}
+              name={'check-console-expose'}
+              id={'check-expose-console'}
+              onChange={setConsoleExpose}
+            />
```
Msarawan, line 112: @lavocatt , The checkbox for "Expose Console" is pre-checked in the UI. However, it seems selecting or unselecting it doesn't affect functionality. Should we keep it, considering this observation? Also, should we add a mandatory symbol to clarify its importance?

src/configuration/broker-models.tsx
```diff
@@ -919,18 +934,40 @@ export const NamingPanel: FC<NamingPanelProps> = ({ initName, uniqueSet }) => {
...
+          <Button
+            key="confirm"
+            variant="primary"
+            onClick={handleNewName}
+            isDisabled={validateStatus !== ValidatedOptions.success}
+          >
+            Confirm
+          </Button>,
+          <Button key="cancel" variant="link" onClick={() => setIsOpen(false)}>
+            Cancel
```
Msarawan, line 967: Here also use the Translation to translate both button text.

src/configuration/console-config.tsx
lavocatt: I don't understand, what do you mean by "pre-checked in the UI", the value isn't checked by default. Checking the value isn't mandatory either. Because the user can leave it unchecked if necessary.

gaohoward: Yeah we leave it as checked by default. Otherwise our plugin cannot access the jolokia outside the openshift (in dev mode, in production the console/jolokia is accessed via the service)

lavocatt: Ah it was in the console config, sorry for that guys, I was looking at the acceptor config..

lavocatt:
Here's the new look 
image
image
image


image

image

Msarawan:
> Here's the new look image image image
> 
> image
> 
> image

Looks good!! nice work Thomas...

just a one thought from me when we click on Trash icon to delete the acceptors it's deleting directly so instead that can we add a pre-confirm delete modal there aswell?

Msarawan (dismissed):
LGTM! Good work Thomas.

lavocatt:
> Looks good!! nice work Thomas...
> 
> just a one thought from me when we click on Trash icon to delete the acceptors it's deleting directly so instead that can we add a pre-confirm delete modal before deleting the acceptor? can we add that in a separate PR?

That's a good idea! I'll add a confirmation modal.

src/configuration/console-config.tsx
lavocatt: As discussed on slack we will use a tool to extract the translations for us. I'm resolving the conversation.

src/configuration/console-config.tsx
lavocatt: I'm adding the required field :slightly_smiling_face:

Msarawan: Thanks!

lavocatt:
I have added the confirmation modal for the deletion:

image

image
 
image

gaohoward (approved):

gaohoward:
LTGM. Very nice layout!

Msarawan (approved):

lavocatt:
Thanks for the reviews!

PR #188, opened 2024-06-20
[#177] custom ui element for other parameters

### LINKED ISSUE #177: Form editor: acceptor's other parameters not working
To reproduce

1. Click 'Create Broker` to open up form editor, create an acceptor
2. type something in 'Other parameters' field and nothing appears.

lavocatt:
Other parameters are a key value pair. The previous method was relying on a textfield for the user to enter these value as a comma separated list of properties. Now there's a ui component for it showing keys and values in separated text fields. The user can add a new entry, edit it at will and then update or delete it.

[Screencast from 2024-06-20 16-32-19.webm](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/08ec074b-9a9b-40c6-aa79-fa748ef3b46e)

fixes #177

### DIFF
--- a/src/brokers/utils/add-broker.test.ts
+++ b/src/brokers/utils/add-broker.test.ts
@@ -186,7 +186,10 @@ describe('test the creation broker reducer', () => {
       operation: ArtemisReducerOperations.setAcceptorOtherParams,
       payload: {
         name: 'acceptors0',
-        otherParams: 'aKey=aValue,bKey=bValue',
+        otherParams: new Map<string, string>([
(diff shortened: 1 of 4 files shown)


RoddieKieley (commented):
While I haven't tested the code directly and thus can't attest to it's use in operation, having this there is a win from a usability point of view as if not for any parameter that we don't cover the user has to go to the yaml view to enter it. As there's quite a few acceptor options that are useful depending on the specific use case it's good to provide this.

gaohoward (approved):

PR #189, opened 2024-06-20
[NOISSUE] broker pods list, only show pods

lavocatt:
There was a small issue with the broker pods list, in some instances it could return a list of brokers instead of a list of pods. By removing the call that was getting the list of brokers, this is fixed. I also added empty states for when there's no pods in the broker for whatever reasons.

image

### DIFF
--- a/src/brokers/broker-pods/PodsList.container.tsx
+++ b/src/brokers/broker-pods/PodsList.container.tsx
@@ -2,18 +2,22 @@ import { FC, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
 import {
   K8sResourceCommon,
-  k8sGet,
   useK8sWatchResource,
 } from '@openshift-console/dynamic-plugin-sdk';
-import { AMQBrokerModel, K8sResourceKind } from '../../utils';
+import { K8sResourceKind } from '../../utils';
 import { useTranslation } from '../../i18n';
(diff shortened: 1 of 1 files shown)


src/brokers/broker-pods/PodsList.container.tsx
```diff
@@ -79,7 +64,33 @@ const PodsContainer: FC = () => {
...
+        {brokerPods.length === 0 && (
+          <EmptyState>
+            <EmptyStateIcon icon={SearchIcon} />
+            <Title size="lg" headingLevel="h4">
+              No results found. Check the status of the deployment.
+            </Title>
+            <EmptyStateBody>No results match.</EmptyStateBody>
+          </EmptyState>
+        )}
+        {loadError && (
```
Msarawan, line 84: Hello @lavocatt , I noticed that when there's a load error, both the loading and error states are being displayed simultaneously. It would be more user-friendly to display only the load error state directly right. Could you please update the conditions to ensure only the error state is shown in such cases.

Screenshot from 2024-06-21 12-39-17

lavocatt: Good catch thanks! I'll fix this

lavocatt: Should be fixed now :)

Issue #193, opened 2024-06-21
Openshift 4.15: Dev build and prod build are diverging in behavior


lavocatt:
There is something going on with the production build that we need to understand and hopefully fix before the demo.

## On the production build it is impossible to list the broker pods of a deployed broker
![image](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/322a58cf-b11c-4b57-8a72-9a1d0178b4c2)

Even though the pod is running fine and the status says "ok":
![image](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/48382226-5421-40bf-825a-c6e86e880c2a)

## On the production build, upon creating a broker, the namespace selection is broken
![image](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/19ac7774-1e2a-45e8-9517-f9b8278202a9)
Even setting the namespace manually doesn't update the field!

lavocatt:
My intuition is telling me that this might be because when we work with the dev environment we are running the version 0.0.21 of the dynamic plugin sdk, but when we run in prod, the dynamic plugin sdk is running the version 1.x.x

I'm trying to update the dynamic sdk version on my branch to see if I can get to reproduce these issues in dev mode. That would be a good start.

lavocatt:
Intuition was correct! I can reproduce the bugs in dev now. I'm opening a draft PR to track the progress on the matter.

lavocatt:
For 4.15, react-router also needs to get updated.

lavocatt:
I've been running tests against 4.12 and our production build is working fine:
![image](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/9802a0f9-596b-49a1-a3ce-65044df9cfcb)

I will try with 4.13 and 4.14.

lavocatt:
After a more thorough exploration it seems like this is due to the react-router library. We're using an outdated version of it. We must upgrade it.

PR #194, opened 2024-06-21
[#193] update the sdk to version 1.3 for Openshift 4.15

lavocatt:
There is a problem with our current setup, it was made to function on a very outdated version of the dynamic plugin sdk (0.0.21). Running on dev mode this wasn't an issue since the entire console was under our control and matching the version number.

But once we build to prod, this is a different matter, the dynamic plugin version that is running is the one from the cluster and has to match. Therefore some of the calls we are running don't work at all.

This is the first step to debug the situation. Following commits will address the diverse bugs we are experiencing in the UI.

See #193


lavocatt:
Closing in favor or #228

Issue #164, opened 2024-06-10
There should not be `alert()` in the codebase


lavocatt:
There are better ways to notify the user instead of alerts.
Let's have a cleanup of the code base and fix this up.
Also adding a linter rule to forbid the use of alerts would be good.

Msarawan:
Closed as completed.

PR #195, opened 2024-06-21
[#164] remove all alert() from the codebase

Msarawan:
Removed all alert() calls from the codebase and replaced them with other user notification methods. Added a linter rule to forbid the use of alerts.

Screenshot from 2024-06-24 13-50-14


lavocatt (approved):
Good job!

PR #196, opened 2024-06-24
[NOISSUE] fix: Refactor Modal component

Msarawan:
Refactored Acceptor and Broker delete Modal component.

Screenshot from 2024-06-24 14-17-13

Screenshot from 2024-06-24 14-21-23

### DIFF
--- a/locales/en/plugin__activemq-artemis-self-provisioning-plugin.json
+++ b/locales/en/plugin__activemq-artemis-self-provisioning-plugin.json
@@ -135,5 +135,9 @@
   "try_again": "Try again",
   "login_failed": "Login Failed",
   "login_failed_message": "Token invalid. Please check your credentials and try again.",
-  "show_cert_info": "only support tls format secret from cert-manager"
+  "show_cert_info": "only support tls format secret from cert-manager",
+  "delete_instance_title": "Permanently delete the {{subject}}",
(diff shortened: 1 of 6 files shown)


lavocatt (changes requested):
The confirmation modal to delete the acceptor must also work for connectors:
image
The issue in the current proposal is that that text doesn't adapt to what is supposed to get deleted. This is a regression in behavior, can you fix it please?

Msarawan:
> The confirmation modal to delete the acceptor must also work for connectors: image The issue in the current proposal is that that text doesn't adapt to what is supposed to get deleted. This is a regression in behavior, can you fix it please?

@lavocatt , means we need to display the selected acceptor/connectors name in the description message right as we were displaying in the brokers?

Msarawan:
> > @lavocatt , means we need to display the selected acceptor/connectors name in the description message right as we were displaying in the brokers?
> 
> It means that the title should be: `Permanently delete the acceptor` when the user deletes an acceptor. And `Permanently delete the connector` when the user deletes a connector.
> 
> Look at the what the code was doing before your changes: image
> 
> The `subject` of the deletion was passed as a parameter. And was used in the title. Is it clearer?

yes Got it. i will update the PR. thanks!

lavocatt (approved):
Thanks for these changes! Good work!

PR #201, opened 2024-06-24
Replace Unicode with PatternFly Icon for View Certification Details Button

Msarawan:
The certification details button in the UI currently uses the Unicode character `\u2687`. replaced this unicode with a Patternfly icon to improve the user interface.

Screenshot from 2024-06-25 00-13-41


lavocatt:
I'm not opposed to the idea to change the button as it looks more semantically spot on than the unicode one. @howardgao wdyt?

lavocatt (changes requested):
There's an issue on the alignment now:
image

The size of the information circle set to MD most probably forces the button to render to widely.

You should try with a smaller size.

lavocatt:
Also there is something going on with your commits. Two questions:
* Why is there two commits?
* Why is the second commit undoing most of what's done in the first commit?

gaohoward:
> I'm not opposed to the idea to change the button as it looks more semantically spot on than the unicode one. @howardgao wdyt?

yes that's definitely better.

Msarawan:
> Also there is something going on with your commits. Two questions:
> 
> * Why is there two commits?
> * Why is the second commit undoing most of what's done in the first commit?

yeah, actually i was checkout with previous branch instead of checkout from main branch that's what there are two commits. let me update the commits now.

gaohoward (approved):

lavocatt (approved):

PR #212, opened 2024-06-26
[#211] remove storybook

### LINKED ISSUE #211: storybook: if it's not used, let's remove it.
I'm taking a look at storybook. There's an old issue mentionning that storybook doesn't build #24 
I've been testing the command and indeed it fails.
It's been a year since the bug was opened and it doesn't seem like story book is a critical element of our development process, since we were able to work without it.

I propose to remove it from our dependencies and to clean the associate code.
Wdyt?

lavocatt:
It's been more than a year that our storybook is down. The issue raised by the team at that time on the dyanmic plugin sdk has gone stale https://github.com/openshift/dynamic-plugin-sdk/issues/216. Which means that we can't hope for a resolution unless we actively engage with them again on the subject.

Given the fact that there are other ways to test the code (react-testing-library) and the fact that none of the current members of the team know how to operate/leverage storybook. Let's just remove it.

This removes a lot of dependencies to external libraries and makes us resemble more of a vanilla plugin, which in turn means easier maintenance down the line.

fixes #211


Msarawan (dismissed):
LGTM!

lavocatt:
The PR #221 got merged. We're all set to merge this one.

Msarawan (approved):
LGTM! Good work Thomas.

PR #215, opened 2024-06-26
release: pushing to quay was failing

lavocatt:
The script was crashing most probably because of this:

```
Warning:
"quay.io/***/quay.io/***/activemq-artemis-self-provisioning-plugin" does not seem to be a valid registry path. The registry path should not contain more than 2 slashes. Refer to the Inputs section of the readme for naming image and registry.
```

The build from the ci works however, I'm taking the parametrisation from this workflow to apply it to the release workflow.


RoddieKieley (approved):
Looks straightforward enough

Issue #210, opened 2024-06-26
code structure: there seems to be too many `utils` folders


lavocatt:
IMO there are too many utils folder in the code base. The issue with `utils` is that it is a generic name and does not help much to find what's inside.

For instance, the reducer for the broker creation form state management lives under `src/brokers/utils/add-broker.ts`. We should be able to find a better name and placement.

lavocatt:
- [ ] functions in `utils/condition-utils.ts` should get moved near the `ConditionModals` and types in `k8s.ts` should get moved near `ConditionModals`
- [ ] `utils/constants.ts` should get moved to `src/constants`
- [ ] `utils/models.ts` should get moved to `k8s/models.ts`
- [ ] `utils/prometheus.ts` should get moved near the `Metrics`
- [ ] `utils/types.ts` should get moved in `k8s/types.ts`

Issue #207, opened 2024-06-26
code structure: refine the component hierarchy for `configuration`


lavocatt:
The `src/configuration` folder contains components, but they are not following the common hierarchy of the project.
In turn, this means it is harder to find where is a component defined.

PR #221, opened 2024-07-02
Refactoring & cleanup: sorting out the components in their respective hierarchy

lavocatt:
There was a bit of an issue in the code base. Related to #207 and #210. That made components hard to find.

This PR acts as a big cleanup phase to sort out all the components and to puts them back in the hierarchy where they belong. For instance, `Metrics` was only used in the `BrokerDetails` it's been moved there

Now we have the `shared-components` directory that is only containing true shared components now. Shared components are used by different entry points (for instance the FormView is used by the add-broker and edit-broker)

I've added a `reducers` folder to contain all the reducers we're going to work on. For now we've got one that is matching 7.12 but that way we can more easily add new ones.

All the components that were under configuration have been unpacked as childs of the FormView shared-component.

A nice followup of this cleanup would be to remove all the `index.ts` files that are a bit cluttering the space without adding much to the table.

Sorry for the huge PR, there's not really a good way to do these kind of cleanups without putting a lot of changes at the same time I'm afraid.


src/brokers/broker-details/components/Overview/Metrics/components/MetricsActions/MetricsActions.tsx
Msarawan, line 1: Hello @lavocatt , I noticed that you already created a commit for moving the metrics folder into the overview component. I think it might be better to include the move of the DropdownWithToggle component from the shared components to the metrics folder in the same commit and update the commit message accordingly. This way, we can avoid creating an extra commit.

Just a suggestion! wdyt?

lavocatt: Sure! That's a good idea :slightly_smiling_face:

lavocatt: I've reduced the number of commits as requested, and even a little more.

Msarawan (approved):
The refactoring looks good. Nice work, Thomas!

gaohoward (approved):

PR #228, opened 2024-07-08
[#193] update the configuration to match the console template

### LINKED ISSUE #193: Openshift 4.15: Dev build and prod build are diverging in behavior

lavocatt:
The configuration got imported from [https://github.com/openshift/console-plugin-template at commit
2b4dbe4b4313e258be1754815614cf7d47f45c11](https://github.com/openshift/console-plugin-template/tree/2b4dbe4b4313e258be1754815614cf7d47f45c11)

The latest patternfly4 compatible versions where recovered from:
https://github.com/openshift/console/blob/649af93d7cb60e7cd9007c4a9a69ef74ec791680/frontend/package.json#L141

The package.json got sorted with npx sort-package-json for easier
comparisons.

Hack:
For some reason, there's something strange happening with victory-core.
If we remove this old configuration from the package.json file:

```
  "resolutions": {
    "victory-core": "<36.9.0"
  },
  "overrides": {
    "victory-core": "<36.9.0"
  },
```

Then there are crashes running the metrics test suite (at the patternfly
level). If we put it back, the UI won't render the metrics anymore.

What is more surprising is that putting this section back, running yarn,
removing it and running yarn once again **fixes both issues**. That's
the reason why I'm leaving that piece of documentation in the commit, in
case we need it back at some point.

What we most probably need to do is to upgrade ASAP to patternfly 5.
Fixes #193


lavocatt:
A possible followup for this PR: update patternfly to version 5

.devcontainer/Dockerfile.console
```diff
@@ -9,6 +9,7 @@ ENV OC_PLUGIN_NAME=$OC_PLUGIN_NAME
 USER root
 CMD eval "oc login $OC_URL -u $OC_USER -p $OC_PASS --insecure-skip-tls-verify" && \
     /opt/bridge/bin/bridge -public-dir=/opt/bridge/static \
+    -i18n-namespaces plugin__$OC_PLUGIN_NAME \
```
Msarawan, line 12: Hello @lavocatt , could you please explain the purpose of adding `i18n-namespaces` here? is it added for translation purpose.

lavocatt: That is from the console plugin template. I've basically imported all the configuration from https://github.com/openshift/console-plugin-template/commit/2b4dbe4b4313e258be1754815614cf7d47f45c11

Msarawan: Got it. Thanks @lavocatt

package.json
```diff
@@ -2,131 +2,127 @@
...
-    "swagger-routes-express": "^3.3.2",
-    "validator": "^13.11.0",
-    "yaml": "^2.4.5"
+  "lint-staged": {
+    "*.{js,ts,tsx}": [
+      "yarn run pretty-quick --staged",
+      "eslint --fix"
+    ]
   },
   "devDependencies": {
```
Msarawan, line 45: Hello @lavocatt , is there any reason for moving all dependencies into the devDependencies section? Initially, I separated them to distinguish between dev and prod dependencies after discussing with Howard. so just curious to know.

Msarawan: Because i referred some other projects and modified like that way. in our code also by separating both dev-dependcies and prod dependencies.

lavocatt: This is from a discussion I had with @vojtech who is one of the maintainer from the dynamic plugin sdk. He told me that we shouldn't separate dependencies and devDependencies and instead to have everything in devDependencies. He said that the only reason why we would need dependencies is if the plugin would get imported by another one and published on npm.

Btw, I'm not sure neither if that is the correct way of doing things, it's just what the ui-extensibility team might want to see in our project when we request assistance from them.

Msarawan: Oh okay, Thanks for the clarification Thomas!

Msarawan (approved):
Changes LGTM. Good work Thomas!

Issue #240, opened 2024-07-15
split the api-server


lavocatt:
move the api-server to its own repository.

PR #241, opened 2024-07-15
[#240] move api server to its own repo

lavocatt:
This commit removes the code from the api-server. The new repository where the code lives now is
https://github.com/artemiscloud/activemq-artemis-jolokia-api-server.

The Dockerfile got synchronized back to the one from the console-plugin-template at commit 2b4dbe4b4313e258be1754815614cf7d47f45c11

The deploy/ folder content has been synchronized back to its state before introducing the api-server in the project. At commit fc70c0ee0ad491a2bb5efb9c23d9cf1691cebde8.

The only difference left is that the plugin relies on the activemq-artemis-jolokia-api-server proxy service to be there:

```
  proxy:
    - type: Service
      alias: api-server-service
      endpoint:
        type: Service
        service:
          name: activemq-artemis-jolokia-api-server
          namespace: activemq-artemis-jolokia-api-server
          port: 9443
```

Follow the updates in the Readme for more information about how to develop and deploy now.

This work is linked to the PR that adds the jolokia-api-server code to its own repo there
https://github.com/artemiscloud/activemq-artemis-jolokia-api-server/pull/4


Dockerfile
```diff
@@ -1,67 +1,17 @@
-FROM registry.access.redhat.com/ubi8/nodejs-20:latest AS BUILD_IMAGE
-
-### BEGIN REMOTE SOURCE
```
Msarawan, line 3: @lavocatt , We need all these changes added within the BEGIN to END remote source. because i added those lines to keep sync with the midstream and downstream repos. For now, could you please revert these changes?

Dockerfile
```diff
@@ -1,67 +1,17 @@
...
-
+FROM registry.access.redhat.com/ubi8/nodejs-16:latest AS build
 USER root
+RUN command -v yarn || npm i -g yarn
 
-## Set directory
-RUN mkdir -p /usr/src/
-RUN cp -r $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app /usr/src/
+ADD . /usr/src/app
 WORKDIR /usr/src/app
```
Msarawan, line 6: same here also. could you please revert the changes.

Dockerfile
```diff
@@ -1,67 +1,17 @@
...
-# Use the COPY instruction only to copy files to the container path $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app
-ARG REMOTE_SOURCES_DIR=/tmp/remote_source
-RUN mkdir -p $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app
-WORKDIR $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app
-# Copy package.json and yarn.lock to the container
-COPY package.json package.json
-COPY yarn.lock yarn.lock
-ADD . $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app
-RUN command -v yarn || npm i -g yarn
-### END REMOTE SOURCE
```
Msarawan, line 14: @lavocatt , We need all these changes added within the BEGIN to END remote source. because i added those lines to keep sync with the midstream and downstream repos. For now, could you please revert these changes?

Dockerfile
```diff
@@ -1,67 +1,17 @@
...
+FROM registry.access.redhat.com/ubi8/nodejs-16:latest AS build
 USER root
+RUN command -v yarn || npm i -g yarn
 
-## Set directory
-RUN mkdir -p /usr/src/
-RUN cp -r $REMOTE_SOURCES_DIR/activemq-artemis-self-provisioning-plugin/app /usr/src/
+ADD . /usr/src/app
 WORKDIR /usr/src/app
+RUN yarn install && yarn build
```
Msarawan, line 7: could you update here with the title in comment lines before the Command. i.e
```
## Install dependencies and Build application
RUN yarn install && yarn build
```

api-server/openapi.yml
```diff
@@ -13,12 +13,12 @@ info:
     to get access to the [management api](https://github.com/apache/activemq-artemis/blob/main/docs/user-manual/management.adoc#the-management-api) of a broker instance via
```
gaohoward, line 13: this openapi.yml need to be sync with the one on api-server repo.
When we do code-gen we can probably use an url , something like

npx --package @7nohe/openapi-react-query-codegen@0.4.1 openapi-rq -i **https://raw.githubusercontent.com/artemiscloud/activemq-artemis-jolokia-api-server/main/openapi.yml** -c fetch -o src/openapi/jolokia && yarn pretty-quick && yarn eslint ./src/openapi --fix",
    "copy-config"

gaohoward (changes requested):

Dockerfile
lavocatt: Can you point me to these repos please?
What is the purpose of these lines exactly?

api-server/openapi.yml
lavocatt: I've thought about that to, but I think it is safer to have a hard copy within the self-provisioning-plugin. There are a couple of reasons why:

* The CI is testing that no manual changes are introduced on the generated code by running the codegen command. If the openapi definition on the `jolokia-api-server` project changes compared to the one that was used to generate the api, it would fail the ci
* There might be a time when the self-provisioning-plugin will lag behind the jolokia-api-server. For instance we could introduce a breaking change on the api-server side that takes time to integrate on the self-provisioning-plugin.

The drawback is that we need to copy the file and open a PR when the openapi definition changes.
We could, instead of copying the file use a commit hash in the url to specify the version we are compatible with.
That wouldn't change the fact that we would need to open a PR to have the self-provisioning-plugin integrate the latest changes, but that could make it more clear which version the codegen is targeting.

Dockerfile
lavocatt: I'm not sure the comment is really necessary as it only paraphrases the commands that executed under, which are install and build. Comments should always bring something more to the table than just stating what is done.

Dockerfile
Msarawan: midstream self-provisioning-plugin repo: https://github.com/rh-messaging/activemq-artemis-self-provisioning-plugin

Dpwnstream self-provisioning-plugin repo: 
https://pkgs.devel.redhat.com/git/containers/amq-broker-self-provisioning-plugin

When I was working on automation for the self-provisioning-plugin during the downstream build to fix issues with the Dockerfiles, I synced all upstream->midstream->downstream Dockerfiles.

api-server/openapi.yml
```diff
@@ -13,12 +13,12 @@ info:
     to get access to the [management api](https://github.com/apache/activemq-artemis/blob/main/docs/user-manual/management.adoc#the-management-api) of a broker instance via 
     its [jolokia](https://jolokia.org/) endpoint.
 
-    ### How to add new apis to the api-server
+    ### How to add new apis to the src
 
     The api server uses [openapi](https://www.openapis.org/) to define
-    its apis. All the apis are defined in the file [openapi.yml](api-server/config/openapi.yml).
+    its apis. All the apis are defined in the file [openapi.yml](src/config/openapi.yml).
```
Msarawan, line 19: @lavocatt , in the codebase the api-server path is still pointing to `api-server/openapi.yml` instead `src/config/openapi.yml`. could you please update this path in the src.

start-console-tls.sh
```diff
@@ -45,13 +45,13 @@ echo "Console Platform: $CONSOLE_IMAGE_PLATFORM"
 if [ -x "$(command -v podman)" ]; then
     if [ "$(uname -s)" = "Linux" ]; then
         # Use host networking on Linux since host.containers.internal is unreachable in some environments.
-        BRIDGE_PLUGINS="${PLUGIN_NAME}=https://localhost:9443"
+        BRIDGE_PLUGINS="${PLUGIN_NAME}=https://localhost:9444"
```
Msarawan, line 48: Why did you change the localhost port from `9443 to 9444`? 
Could you please explain the reason for this change?

lavocatt: Sure, the reason is that the `api-server` starts on 9443 on another process now, so this port is already taken on the machine. The bridge has to use another one. I've selected 9444 for that purpose. The value itself doesn't really matter, since it's only used for the bridge (it's not going to change the url your using to access the tls console).

Dockerfile
lavocatt: All good now, we've discussed offline. The reason why this block is needed is because we only keep the Dockerfile in downstream and code gets injected at build time by another tool. So the upstream Dockerfile has to be compatible with that workflow. I'll add these lines back.

Msarawan (commented):
Hello @lavocatt , on the `Pods List page` and the `Broker Details page`, clicking on both `Edit broker` and `Delete broker` buttons does not open the editors and the delete modal pages. Instead, the below error message is displayed.
Screenshot from 2024-07-19 21-14-51

api-server/openapi.yml
gaohoward: ok, let keep the way it is now until a better solution is found.

gaohoward (approved):

gaohoward:
I approved on my side leaving @Msarawan for the issues she found.

lavocatt:
> Hello @lavocatt , on the `Pods List page` and the `Broker Details page`, clicking on both `Edit broker` and `Delete broker` buttons does not open the editors and the delete modal pages. Instead, the below error message is displayed. Screenshot from 2024-07-19 21-14-51

Thanks a lot for catching this!
While upgrading the console to `1.4.0` the `toggle-kebab` was removed. We also don't need the `onFocus` function anymore.
It's fixed. can you have another look at the PR please @Msarawan ?

Msarawan (approved):
LGTM! Nice work Thomas.

PR #248, opened 2024-07-26
Upon adding a new acceptor/connector update the port number to unique

### LINKED ISSUE #204: Upon adding a second acceptor the second acceptor should have a different port number
When the user adds a second acceptor, the two acceptors have the same port number which fails validation.
To fix: adding a nth acceptor should mean that the port of the acceptor should be 5555+n instead of a static 5555.

Add also validation on the ports fields to ensure uniqueness.
Check that this doesn't collide with connectors as well.

Msarawan:
When user adds a new acceptor or connector updated port number to unique instead of a static 5555.
and also added unit tests to verify the logic.

fixes #204

### DIFF
--- a/src/reducers/7.12/reducer.ts
+++ b/src/reducers/7.12/reducer.ts
@@ -1125,6 +1125,42 @@ const generateUniqueName = (prefix: string, existing: Set<string>): string => {
   return newName;
 };
 
+const generateUniquePort = (cr: BrokerCR): number => {
+  const acceptorSet = listConfigs(
+    ConfigType.acceptors,
+    cr,
+    'set',
+  ) as Set<string>;
+  const connectorSet = listConfigs(
+    ConfigType.connectors,
+    cr,
+    'set',
+  ) as Set<string>;
+
+  const basePort = 5555;
+  if (acceptorSet.size === 0 && connectorSet.size === 0) {
+    return basePort;
+  }
+
+  let maxPort = basePort;
+
+  acceptorSet.forEach((name) => {
+    const port = getConfigPort(cr, ConfigType.acceptors, name);
+    if (port > maxPort) {
+      maxPort = port;
+    }
+  });
+
+  connectorSet.forEach((name) => {
+    const port = getConfigPort(cr, ConfigType.connectors, name);
+    if (port > maxPort) {
+      maxPort = port;
+    }
+  });
+
+  return maxPort + 1;
+};
+
 const addConfig = (cr: BrokerCR, configType: ConfigType) => {
   const acceptorSet = listConfigs(configType, cr, 'set') as Set<string>;
 
@@ -1135,7 +1171,7 @@ const addConfig = (cr: BrokerCR, configType: ConfigType) => {
       name: newName,
       protocols: 'ALL',
       host: 'localhost',
-      port: 5555,
+      port: generateUniquePort(cr),
     };
     if (!cr.spec.connectors) {
       cr.spec.connectors = [connector];
@@ -1146,7 +1182,7 @@ const addConfig = (cr: BrokerCR, configType: ConfigType) => {
     const acceptor = {
       name: newName,
       protocols: 'ALL',
-      port: 5555,
+      port: generateUniquePort(cr),
     };
     if (!cr.spec.acceptors) {
       cr.spec.acceptors = [acceptor];
@@ -1732,19 +1768,18 @@ export const getConfigPort = (
   configType: ConfigType,
   configName: string,
 ): number => {
-  if (configType === ConfigType.connectors) {
-    const connector = getConnector(brokerModel, configName);
-    if (connector?.port) {
-      return connector.port;
-    }
-  }
   if (configType === ConfigType.acceptors) {
     const acceptor = getAcceptor(brokerModel, configName);
     if (acceptor?.port) {
       return acceptor.port;
     }
+  } else if (configType === ConfigType.connectors) {
+    const connector = getConnector(brokerModel, configName);
+    if (connector?.port) {
+      return connector.port;
+    }
   }
-  return 5555;
+  return -1;
 };
 
 export const getConfigHost = (
--- a/src/reducers/7.12/reducer.test.ts
+++ b/src/reducers/7.12/reducer.test.ts
@@ -213,6 +213,28 @@ describe('test the creation broker reducer', () => {
     );
   });
 
+  it('should assigns unique ports to each new acceptor added', () => {
+    const initialState = newArtemisCRState('namespace');
+
+    // Add the first acceptor
+    let newState = artemisCrReducer(initialState, {
+      operation: ArtemisReducerOperations.addAcceptor,
+    });
+    expect(newState.cr.spec.acceptors[0].port).toBe(5555);
+
+    // Add a second acceptor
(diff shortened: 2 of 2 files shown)


src/shared-components/FormView/BrokerProperties/ConfigurationPage/AcceptorsConfigPage/AcceptorConfigSection/AcceptorConfigPage/AcceptorConfigPage.tsx
```diff
@@ -176,6 +178,10 @@ export const AcceptorConfigPage: FC<AcceptorProps> = ({
   ];
 
   const { t } = useTranslation();
+  useEffect(() => {
+    setPort(initialPort);
+  }, [initialPort]);
+
```
lavocatt, line 184: Please don't use `useEffect` for this, refer to https://react.dev/learn/you-might-not-need-an-effect for more details about why this isn't a good pattern.

src/reducers/7.12/reducer.ts
```diff
@@ -1698,23 +1710,63 @@ export const getConnector = (cr: BrokerCR, name: string) => {
...
+      if (connector.port) {
+        usedPorts.push(connector.port);
+      }
+    });
+    const newPort = generatePortNumber(usedPorts);
+    //console.log(`Generated new port for connector ${configName}: ${newPort}`);
+    return newPort;
   }
+
+  //console.log(`default port 5555 for configName: ${configName}`);
```
lavocatt, line 1769: There should not be commented code lines in a PR. Can you clean them all please?

src/reducers/7.12/reducer.ts
```diff
@@ -1698,23 +1710,63 @@ export const getConnector = (cr: BrokerCR, name: string) => {
...
+    const connectors = getAllConnectors(brokerModel);
+    const usedPorts: number[] = [];
+    connectors.forEach((connector) => {
+      if (connector.port) {
+        usedPorts.push(connector.port);
+      }
+    });
+    const newPort = generatePortNumber(usedPorts);
+    //console.log(`Generated new port for connector ${configName}: ${newPort}`);
+    return newPort;
```
lavocatt, line 1766: The function `getConfigPort` is a getter function, it's purpose it to give an easy access to some data we care about. It should not be used to generate a port number.

Can you explain me what was your plan for this section of the code? So that we can find a better way to integrate it.

src/reducers/7.12/reducer.ts
```diff
@@ -298,10 +298,18 @@ interface DeletePEMGenerationForAcceptorAction
...
+    port: number;
+  };
 }
 
 interface AddConnectorAction extends ArtemisReducerActionBase {
   operation: ArtemisReducerOperations.addConnector;
+  payload: {
+    name: string;
+    port: number;
+  };
```
lavocatt, line 312: By asking the calling code to provide the `name` and `port` to the Add connector action, we're externalizing the logic of producing a unique port number to the UI code. That's making us diverging from the primary goal of the reducer: centralizing the update logic.

What we want to have instead is a addConnector function able to search for a free port.

src/reducers/7.12/reducer.ts
```diff
@@ -1698,23 +1710,63 @@ export const getConnector = (cr: BrokerCR, name: string) => {
...
+};
+
+export const generatePortNumber = (usedPorts: number[]): number => {
+  let port = 5555;
+  while (usedPorts.includes(port)) {
+    port += 1;
+    //console.log('increament port', port);
+  }
+  return port;
+};
```
lavocatt, line 1724: There's no need to work with a secondary list of `usedPorts` we can work in place instead. To get a new free port:
Find the biggest port number in all the acceptors, connectors and the console. Then just add +1 to it.

src/reducers/7.12/reducer.ts
```diff
@@ -298,10 +298,18 @@ interface DeletePEMGenerationForAcceptorAction
 
 interface AddAcceptorAction extends ArtemisReducerActionBase {
   operation: ArtemisReducerOperations.addAcceptor;
+  payload: {
+    name: string;
+    port: number;
+  };
```
lavocatt, line 304: same comment as for connectors.

src/shared-components/FormView/BrokerProperties/ConfigurationPage/AcceptorsConfigPage/AcceptorsConfigPage.tsx
```diff
@@ -33,13 +34,25 @@ export const AcceptorsConfigPage: FC<AcceptorsConfigProps> = ({ brokerId }) => {
...
     }
     if (configType === ConfigType.connectors) {
+      const newPort = getConfigPort(
+        cr,
+        configType,
+        `connector-${configs.length}`,
+      );
       dispatch({
         operation: ArtemisReducerOperations.addConnector,
+        payload: { name: `connector-${configs.length}`, port: newPort },
```
lavocatt, line 55: See, we already can see some logic externalized from the reducer to the frontend code. A hint for you: the correct solution should imply 0 changes to the UI side of the code.

lavocatt (changes requested):
You've made a good first attempt at solving the issue. But this is a hard one. It involves advanced concepts about where the logic should leave and where it shouldn't. I'm expecting that we will have many back and forth on this one, but don't get discouraged, you'll succeed, no doubt.

What to keep in mind when you're working a second iteration of your contribution:

- The reducer owns the logic
- The UI  components are dispatching events, they can't compute what is the free port
- Getters are purely a way to get to existing data, they can't generate some on the fly.
- Only a solution where the UI components are left untouched will be considered valid.

Have a read on react documentation about reducers, and try out the tutorials they provide, it's good learning: https://react.dev/learn/extracting-state-logic-into-a-reducer

lavocatt (commented):
That's a very good improvement. Now the two things left to do are:

* improve the algorithm for generating a unique port number
* add tests to make sure this algorithm does what it says it's doing

lavocatt (commented):
That's a very good algorithm you've wrote.

Can you add a test that mixes acceptors and connectors please?

like it would add an acceptor, then a connector, then set the port for the first acceptor and then add another connector. With checking the expected value for every acceptors and connectors.

That would complement well the new tests you've already added.

lavocatt (approved):
This is perfect! Very good job.

Msarawan:
> This is perfect! Very good job.

Thank you...

gaohoward:
@Msarawan @lavocatt Sorry a bit late but just want to add a little note here:
The port value is not necessarily unique between acceptors and connectors. For example when an acceptor port is 5555, the connector's port can also be 5555 too.

lavocatt:
@gaohoward can you create an issue with this please? So we can fix it in a followup PR? thanks

Msarawan:
> @Msarawan @lavocatt Sorry a bit late but just want to add a little note here: The port value is not necessarily unique between acceptors and connectors. For example when an acceptor port is 5555, the connector's port can also be 5555 too.

ohh okay, will fix this with new PR. thanks!!

gaohoward:
> The port value is not necessarily unique between acceptors and connectors. For example when an acceptor port is 5555, the connector's port can also be 5555 too.

will do.

PR #250, opened 2024-07-29
Add unit tests for all the components in the addresses/address-details  

Msarawan:
Added unit tests for AddressDetailsRow component

### DIFF
--- a/src/addresses/address-details/AddressDetails.component.test.tsx
+++ b/src/addresses/address-details/AddressDetails.component.test.tsx
@@ -1,0 +1,148 @@
+import { fireEvent, render, screen } from '../../test-utils';
+import { AddressDetails } from './AddressDetails.component';
+import { JolokiaAuthentication } from '../../jolokia/components/JolokiaAuthentication';
+import { useJolokiaServiceReadAddressAttributes } from '../../openapi/jolokia/queries';
+import { useJolokiaLogin } from '../../jolokia/customHooks';
+
+jest.mock('../../openapi/jolokia/queries', () => ({
+  useJolokiaServiceReadAddressAttributes: jest.fn(),
+}));
+
+jest.mock('../../jolokia/customHooks', () => ({
+  useJolokiaLogin: jest.fn(),
+  useGetApiServerBaseUrl: jest.fn(),
+}));
+
+const mockUseJolokiaServiceReadAddressAttributes =
+  useJolokiaServiceReadAddressAttributes as jest.Mock;
+const mockUseJolokiaLogin = useJolokiaLogin as jest.Mock;
+
+describe('AddressDetails', () => {
+  const mockData = [
+    { request: { attribute: 'Address' }, value: 'DLQ' },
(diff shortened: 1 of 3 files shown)


lavocatt (commented):
That looks good to me, I'm wondering if there's more we could test?

Msarawan:
> That looks good to me, I'm wondering if there's more we could test?

Yes, within AddressDetails we have 3 components we need to write tests for all 3 components. because it involves logic of query client and AuthContext so we need to check for all the components.

lavocatt:
> > That looks good to me, I'm wondering if there's more we could test?
> 
> Yes, within AddressDetails we have 3 components we need to write tests for all 3 components. because it involves logic of query client and AuthContext so we need to check for all the components.

Can you add these too in the PR please?

src/addresses/address-details/AddressDetails.component.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const { container } = render(
+      <AuthContext.Provider value="mock-token">
+        <AddressDetails name="DLQ" />
+      </AuthContext.Provider>,
+    );
+    const currentPage = container.querySelector(
+      '#pagination-options-menu-top-top-pagination > nav > div:nth-child(3) > input',
+    );
+    expect(currentPage).toHaveValue(1);
+  });
```
lavocatt, line 114: Can you update this test so that it clicks on the next page and asserts that the page number is 2 please?

gaohoward (approved):

src/addresses/address-details/AddressDetails.component.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+      error: null,
+    });
+
+    render(
+      <AuthContext.Provider value="mock-token">
+        <AddressDetails name="DLQ" />
+      </AuthContext.Provider>,
+    );
+
+    expect(screen.getByText('attributes')).toBeInTheDocument();
```
gaohoward, line 61: Here should it be checking the attribute values are empty or absent in case of read failure?

src/addresses/address-details/AddressDetails.component.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+    //initial data rendered
+    expect(screen.getByText('Address')).toBeInTheDocument();
+    expect(screen.getByText('AddressSize')).toBeInTheDocument();
+    expect(screen.getByText('AutoCreated')).toBeInTheDocument();
+    expect(screen.getByText('MessageCount')).toBeInTheDocument();
+    expect(screen.getByText('RoutingTypes')).toBeInTheDocument();
+
+    const searchInput = screen.getByPlaceholderText('search');
+    fireEvent.change(searchInput, { target: { value: 'Message' } });
+
```
lavocatt, line 105: You can insert 

```
    expect(screen.getByText('MessageCount')).toBeInTheDocument();
    expect(screen.queryByText('Address')).not.toBeInTheDocument();
```
    
In here

lavocatt (approved):
Very good job!

PR #255, opened 2024-07-30
Fix typo in openapi.yml description

gemmellr:


lavocatt:
Thanks @gemmellr for fixing the typo! Since we've split the api-server to its own project, this file is a copy of https://github.com/artemiscloud/activemq-artemis-jolokia-api-server/blob/main/src/config/openapi.yml. Can you make the fix over there please? Then we'll import the modification on a second PR.

gemmellr:
Wishing I hadn't bothered now :)

Raised: https://github.com/artemiscloud/activemq-artemis-jolokia-api-server/pull/6

PR #259, opened 2024-08-02
fix: Incorrect Pod Filtering for Brokers Starts with similar prefixes/names 

Msarawan:
- fixed bug with incorrect Pod Filtering for Brokers Starts with similar prefixes/names.
- now, filtering works as expected with selected brokername.

### DIFF
--- a/src/brokers/broker-pods/PodsList.container.tsx
+++ b/src/brokers/broker-pods/PodsList.container.tsx
@@ -36,9 +36,10 @@ export const PodsContainer: FC = () => {
   });
 
   const filterBrokerPods = (pods: K8sResourceCommon[], brokerName: string) => {
+    const regex = new RegExp(`^${brokerName}-ss$`);
     return pods.filter((pod) =>
       pod.metadata?.ownerReferences?.some(
-        (ref) => ref.name.startsWith(brokerName) && ref.kind === 'StatefulSet',
+        (ref) => regex.test(ref.name) && ref.kind === 'StatefulSet',
       ),
     );
   };
--- a/src/brokers/broker-pods/PodsList.container.test.tsx
+++ b/src/brokers/broker-pods/PodsList.container.test.tsx
@@ -1,0 +1,94 @@
+import { render, screen, waitFor } from '../../test-utils';
+import { PodsContainer } from './PodsList.container';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+import { useParams } from 'react-router-dom-v5-compat';
+import { K8sResourceCommon } from '@openshift-console/dynamic-plugin-sdk';
+
+jest.mock('@openshift-console/dynamic-plugin-sdk', () => ({
+  useK8sWatchResource: jest.fn(),
+}));
+
+jest.mock('react-router-dom-v5-compat', () => ({
+  useParams: jest.fn(),
+}));
+
+jest.mock('./components/PodList', () => ({
+  PodsList: jest.fn(() => <div>PodsList Component</div>),
+}));
+
+jest.mock('./components/BrokerPodsBreadcrumb/BrokerPodsBreadcrumb', () => ({
+  BrokerPodsBreadcrumb: jest.fn(() => (
+    <div>BrokerPodsBreadcrumb Component</div>
+  )),
+}));
+
+describe('PodsContainer', () => {
+  const mockUseK8sWatchResource = useK8sWatchResource as jest.Mock;
+  const mockUseParams = useParams as jest.Mock;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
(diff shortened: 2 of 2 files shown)


src/brokers/broker-pods/PodsList.container.tsx
```diff
@@ -36,11 +36,24 @@ export const PodsContainer: FC = () => {
...
+      if (!pod.metadata?.ownerReferences) {
+        return false;
+      }
+
+      return pod.metadata.ownerReferences.some((ref) => {
+        const refName = ref.name.toLowerCase();
+        const refNameWithoutSuffix = refName.endsWith('-ss')
+          ? refName.slice(0, -3)
+          : refName;
+        if (ref.kind !== 'StatefulSet') {
```
gaohoward, line 50: This check can be put in front of the refNameWithSuffix extraction in case it is true the extraction operation will be skipped.

gaohoward:
@Msarawan We could add a simple unit test for the filtering logic.

src/brokers/broker-pods/PodsList.container.tsx
```diff
@@ -36,11 +36,24 @@ export const PodsContainer: FC = () => {
   });
 
   const filterBrokerPods = (pods: K8sResourceCommon[], brokerName: string) => {
-    return pods.filter((pod) =>
-      pod.metadata?.ownerReferences?.some(
-        (ref) => ref.name.startsWith(brokerName) && ref.kind === 'StatefulSet',
-      ),
-    );
+    const BrokerName = brokerName.toLowerCase();
```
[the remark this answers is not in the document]
Msarawan: oh! okay, thanks @gaohoward  for confirming. i will update it.

gaohoward (approved):

src/brokers/broker-pods/PodsList.container.tsx
```diff
@@ -36,11 +36,22 @@ export const PodsContainer: FC = () => {
...
+          return false;
+        }
+        const refName = ref.name;
+        const refNameWithoutSuffix = refName.endsWith('-ss')
+          ? refName.slice(0, -3)
+          : refName;
+        const isMatch = refNameWithoutSuffix === brokerName;
+        return isMatch;
+      });
+    });
```
lavocatt, line 54: While this seems good and looks to work quite well, I think we can improve it by using a regex, something along those lines:

```typescript
    return pods.filter((pod) =>
      pod.metadata?.ownerReferences?.some(
        (ref) => ref.name.match(`^${brokerName}-ss[0-9]$?`) && ref.kind === 'StatefulSet',
      ),
    );
```

src/brokers/broker-pods/PodsList.container.tsx
```diff
@@ -36,11 +36,22 @@ export const PodsContainer: FC = () => {
...
 
   const filterBrokerPods = (pods: K8sResourceCommon[], brokerName: string) => {
-    return pods.filter((pod) =>
-      pod.metadata?.ownerReferences?.some(
-        (ref) => ref.name.startsWith(brokerName) && ref.kind === 'StatefulSet',
-      ),
-    );
+    return pods.filter((pod) => {
+      if (!pod.metadata?.ownerReferences) {
+        return false;
```
lavocatt, line 41: Why has the function shifted from returning a list of pods to a boolean?

lavocatt (changes requested):
Good work, some comments down below

src/brokers/broker-pods/PodsList.container.tsx
Msarawan: with my initial implementation i designed to return a list of pods that are associated with a specific broker by using filter method. here the line40 checks if the pod has ownerReferences. If it doesn’t, the function returns false for the selected pod, excluding it from the filtered results.

but now as i am updating with regex pattern i will remove this condition.

src/brokers/broker-pods/PodsList.container.tsx
gaohoward: Pods name follows the pattern <brokerName>-ss-<oridinal number> . The regex need to have a unit test.

gaohoward (dismissed):

src/brokers/broker-pods/PodsList.container.tsx
Msarawan: > Pods name follows the pattern `[brokerName]-ss-[oridinal number]` . So the owner ref name (statefulset) is `[brokerName]-ss`

Hello @gaohoward , yes that's correct the statefulset has pattern of `[brokerName]-ss` i updated this same approach in my changes.

Msarawan:
Hello @gaohoward and @lavocatt i updated unit tests for `PodsListContainer` component in this PR itself to verify all the states and filtering logic in the `PodsListContainer` works as expected. please take a look.

lavocatt (approved):
Nice work! All looks good to me :)

PR #268, opened 2024-08-12
[#269] remove the ingressDomain field

### LINKED ISSUE #269: Do not display the IngressDomain field in the formview

lavocatt:
The ingress domain can be retrieved from the cluster configuration instead of asking the user to fill in the information themselves. This removes a confusing field from the formview.

This resolves one of the feedback points from @brusdev during the plugin presentation.

fixes #269

### DIFF
--- a/src/brokers/add-broker/AddBroker.container.tsx
+++ b/src/brokers/add-broker/AddBroker.container.tsx
@@ -13,6 +13,7 @@ import {
 } from '../../reducers/7.12/reducer';
 import { AddBrokerResourceValues } from '../../reducers/7.12/import-types';
 import { useNavigate, useParams } from 'react-router-dom-v5-compat';
+import { UseGetIngressDomain } from '../../k8s/customHooks';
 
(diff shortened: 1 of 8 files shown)


Msarawan (commented):
LGTM!! nice work Thomas!

Msarawan:
@lavocatt , just one small suggestion as the ingress domain name is dynamically fetching from  cluster configuration .we can remove the `Ingress Host` input from acceptor configuration also right.
Screenshot from 2024-08-14 19-04-25

lavocatt:
> @lavocatt , just one small suggestion as the ingress domain name is dynamically fetching from cluster configuration .we can remove the `Ingress Host` input from acceptor configuration also right. Screenshot from 2024-08-14 19-04-25

I would prefer to keep the ingress host as this is a valuable field to customize, you can set this value without the preset to setup your ingress exposure of the broker.

Msarawan (approved):

PR #270, opened 2024-08-13
[#217] checking certManager with CRDs

### LINKED ISSUE #217: Production build: UI wrongly says that there is no cert-manager installed
image

I've tried with the community cert-manager and the downstream cert-manager, the result was the same.
But it's possible to create a chain of trust on preset menu:

image

We need to figure out the differences between the two of them.

### LINKED ISSUE #222: Cert-manager checking logic is broken
Currently the plugin verify if a cert-manager is installed is broken. It only looks for deployment of cert-manager in `cert-manager` namespace. If cert-manager is installed in a different namespace the verification doesn't work.

Consider to remove it and find a better way to do the verification.

lavocatt:
Update the cert-manager checking logic to use CRDs instead of checking if cert-manager is deployed in a particular namespace. This is giving better results because cert-manager could have been deployed in many different namespaces.

image

image

fixes #222 
fixes #217

### DIFF
--- a/src/shared-components/FormView/BrokerProperties/ConfigurationPage/AcceptorsConfigPage/AcceptorConfigSection/PresetButton/PresetButton.tsx
+++ b/src/shared-components/FormView/BrokerProperties/ConfigurationPage/AcceptorsConfigPage/AcceptorConfigSection/PresetButton/PresetButton.tsx
@@ -3,6 +3,7 @@ import {
   ButtonVariant,
   Card,
   CardBody,
+  CardFooter,
   CardTitle,
   Form,
   FormFieldGroup,
@@ -24,6 +25,8 @@ import {
   getCertManagerResourceTemplateFromAcceptor,
 } from '../../../../../../../reducers/7.12/reducer';
 import { SelectIssuerDrawer } from '../SelectIssuerDrawer/SelectIssuerDrawer';
+import { useHasCertManager } from '../../../../../../../k8s/customHooks';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
 type PreconfigurationButtonProps = {
   acceptor: Acceptor;
 };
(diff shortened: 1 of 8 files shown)


src/k8s/models.ts
```diff
@@ -68,3 +68,14 @@ export const SecretModel: K8sModel = {
   namespaced: false,
   crd: true,
 };
+
+export const CertManager: K8sModel = {
```
gaohoward, line 72: Does this work for community cert-manager?

src/shared-components/FormView/BrokerProperties/ConfigurationPage/AcceptorsConfigPage/AcceptorConfigSection/PresetButton/PresetButton.tsx
```diff
@@ -79,6 +82,10 @@ const AddPresetModal: FC<AddIssuerAnnotationModalProps> = ({
     getCertManagerResourceTemplateFromAcceptor(cr, initialAcceptor) !==
     undefined;
   const [showCertManagerForm, setShowCertManagerForm] = useState(false);
+  const { hasCertManager, isLoading: isLoadingCertManagerAvailability } =
+    useHasCertManeger();
+  const isCertMangerDependncySatisfied =
```
gaohoward, line 87: typo ? `isCertMangerDependncySatisfied`

src/k8s/models.ts
lavocatt: Good catch, it doesn't 
image

I guess I need also to cover that usecase

lavocatt: Unfortunately the CRD CertManager doesn't exist when the community cert manager is created.

gaohoward: Maybe we can check the presence of the Certificate, Issuer, ClusterIssuer CRDs.

gaohoward (approved):

PR #271, opened 2024-08-13
[#192] support multiple hosts for the PEM case

### LINKED ISSUE #192: PEM certificate generation: support multiple replicas
ATM there aren't correct support for multiple replicas on the reducer. An easy way could be to simply add a host in the patch for every pod in the deployment:

image

This has for effect to spawn two ingresses:

image

Which is functional. However, using this strategy we can't use the ingress round robin for load balancing.

So I'm wondering, is this the right approach?

@brusdev and @gtully I'm pinging you here to have your opinion on the matter.

lavocatt:
Add support for multiple hosts to be exposed via an ingress when the user selects several replicas. Each pod gets exposed and has its own url to get connected to.

The connectivity helper shows the url of the currently selected pod.

fixes #192

### DIFF
--- a/src/reducers/7.12/reducer.ts
+++ b/src/reducers/7.12/reducer.ts
@@ -617,16 +617,19 @@ export const artemisCrReducer: React.Reducer<
       updateNamespace(formState.cr, action.payload);
       break;
     case ArtemisReducerOperations.setReplicasNumber:
-      formState.cr.spec.deploymentPlan.size = action.payload;
+      updateDeploymentSize(formState.cr, action.payload);
       break;
     case ArtemisReducerOperations.incrementReplicas:
-      formState.cr.spec.deploymentPlan.size += 1;
+      updateDeploymentSize(
+        formState.cr,
+        formState.cr.spec.deploymentPlan.size + 1,
+      );
       break;
     case ArtemisReducerOperations.decrementReplicas:
-      formState.cr.spec.deploymentPlan.size -= 1;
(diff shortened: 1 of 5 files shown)


gaohoward (approved):

lavocatt:
@gaohoward I had to rebase the PR, can you give it another check please?

src/reducers/7.12/reducer.ts
```diff
@@ -884,7 +887,31 @@ const updateIngressDomain = (cr: BrokerCR, newName: string) => {
...
+  // when the size changes, some annotations will need an update to
+  // stay in sync
+  if (!cr.spec.acceptors) {
+    return;
+  }
+  if (!cr.spec.resourceTemplates) {
+    return;
+  }
+  cr.spec.acceptors.forEach((acceptor) => {
+    const rt = cr.spec.resourceTemplates.find(
```
Msarawan, line 908: @lavocatt , Good work on this.
just a question here what is mean by variable `rt` here?

lavocatt: `rt` is an abbreviation for `resource templace`.
`const resourceTemplate = cr.spec.resourceTemplates.find(` felt too verbose and redundant, since we're invoking `find` on the list `resourcesTemplates` we know we'll get a `resourceTemplate`, hence the abbreviation.

Msarawan: got it thanks!

Msarawan (approved):
LGTM, Nice work Thomas.

Issue #226, opened 2024-07-05
code quality: rename the type K8sResourceKind to something better


lavocatt:
And also asses its usages. Some looks contradictory.

Msarawan:
> Removing loadash would mean a lot of refactoring for the metrics components. I'm not sure this is worth the effort. We should make sure that we don't use this library for new components as much as possible in the future. Wdyt @Msarawan ? I think we can close this issue.

yeah i agree with your point.

PR #272, opened 2024-08-13
[#226] better naming for generic resource

lavocatt:
The generic resource K8sResourceKind was composed from a BrokerCR and a generic data field. This definition was wrong since it was used to recover generic fields such as CertManager results.

Now it is named K8sResourceCommonWithData and is composed from a K8sResourceCommon plus a generic data field.


Msarawan (approved):
LGTM!

Issue #276, opened 2024-08-21
The deployment size should be allowed to reach the value 0


lavocatt:
A size of 0 is a valid size for the deployment.

PR #282, opened 2024-09-02
The deployment size should be allowed to reach the value 0

Msarawan:
Updated the deployment size should be allowed to reach the value 0.

Screenshot from 2024-09-02 18-30-48

fixes: [#276](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/276)

### DIFF
--- a/src/reducers/7.12/reducer.test.ts
+++ b/src/reducers/7.12/reducer.test.ts
@@ -50,8 +50,8 @@ describe('test the creation broker reducer', () => {
     const newState = artemisCrReducer(initialState, {
       operation: ArtemisReducerOperations.decrementReplicas,
     });
-    // default size is 1 decrementing should result of a size of 1
-    expect(newState.cr.spec.deploymentPlan.size).toBe(1);
+    // default size is 1 decrementing should result of a size of 0
+    expect(newState.cr.spec.deploymentPlan.size).toBe(0);
     // set the number of replicas to 10 before decrementing so that the total
     // number should be 9
     const newState2 = artemisCrReducer(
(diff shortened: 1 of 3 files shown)


lavocatt (changes requested):
This is good, I believe that there were tests testing that the value could not go below 1, they should also be updated :slightly_smiling_face: 

edit: this is what is failing:

image

Msarawan:
> This is good, I believe that there were tests testing that the value could not go below 1, they should also be updated 🙂
> 
> edit: this is what is failing:
> 
> image

yes i am updating the tests.

src/reducers/7.12/reducer.test.ts
```diff
@@ -50,8 +50,8 @@ describe('test the creation broker reducer', () => {
     const newState = artemisCrReducer(initialState, {
       operation: ArtemisReducerOperations.decrementReplicas,
     });
-    // default size is 1 decrementing should result of a size of 1
-    expect(newState.cr.spec.deploymentPlan.size).toBe(1);
+    // default size is 1 decrementing should result of a size of 0
+    expect(newState.cr.spec.deploymentPlan.size).toBe(0);
```
lavocatt, line 54: Can you add a state manipulation in this test to try to decrement under 0 and make sure that the value doesn't go below 0 please?

```
    const newState = artemisCrReducer(initialState, {
      operation: ArtemisReducerOperations.decrementReplicas,
    });
    // decrementing from 0 should leave a value of zero
    expect(newState.cr.spec.deploymentPlan.size).toBe(0);
```

lavocatt (approved):
All good to me! Good work

Msarawan:
> All good to me! Good work

Thank you!

Issue #219, opened 2024-06-28
The UI should be compatible with CRs that are containing empty arrays.


lavocatt:
![image](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/assets/86971992/2193502b-1d86-494d-978c-9bfc0e7f2fdd)

On the left hand side of the screen is a CR where where the user clicked first on add an acceptor then add a connector and then finally deleted them. This had for effect to leave empty arrays in the spec configuration. In turn this makes the UI crash because it tries to perform operations on stuff that doesn't exist. The last bit necessitate more digging.

One way to fix it is to make sure that there are no leftovers in the spec when the user deletes a config.

lavocatt:
@gaohoward has a very good point. what we need to do is
* create a broker with two empty arrays of connectors and acceptors
* navigate to the brokers details page and fix the bugs there
* make sure every pages function correctly under these conditions

PR #284, opened 2024-09-09
fix: UI should be compatible with CRs that are containing empty arrays

Msarawan:
This commit resolves a crash in the `ConnectivityHelper` component caused by calling the reduce() function on an empty array without an initial value. In cases where the `acceptors` array in the CR's spec is empty, the UI previously crashed due to the lack of an initial value in the reduce operation.

The fix ensures that the reduce function uses `false` as the initial value, making the UI compatible with CRs containing empty `acceptors` arrays.

fixes: [#219](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/219)


lavocatt (approved):
That's a very good fix. Thanks

PR #285, opened 2024-09-10
[#208] make @app working properly

### LINKED ISSUE #208: @app is not working properly
The @app tag used in the imports is not working at compile time, but the linter can find the imports.
Something is off in the typescript configuration about this.

Two options: either drop @app or make it compile.

lavocatt:
Webpack and jest needed to be aware of the @app alias to be able to find the related sources.

fixes #208


Msarawan (approved):
LGTM, that's a Good fix!

Issue #218, opened 2024-06-28
When creating a broker, the broker page doesn't refresh automatically


lavocatt:
In turn, this means that the user needs to reload manually the page to see the status evolving.
This is a pain point for the user experience.

PR #286, opened 2024-09-12
fix: Broker page refreshes automatically upon broker creation

Msarawan:
fixed the bug where the broker page refreshes automatically upon broker creation using `useK8sWatchResource` hook.

fixes: [#218](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/218)


lavocatt (commented):
I've just tested it and it addresses the issue perfectly. Good job. It needs a bit of polish with some conflicts to solve and you're good to go.

lavocatt (approved):
LGTM!

PR #287, opened 2024-09-13
[NO-ISSUE] post patternfly5 upgrade on cards

lavocatt:
After the upgrade to pf5 got complete, and after the fix preventing some CSS to load completely was landed in the console, there was still some work to be done to migrate the cards to the new PF5 ways of doing things.
Also a check was left in the cert generation preventing certs to be generated even though cert-manager was installed.


Msarawan (approved):
LGTM, Good fix.

PR #290, opened 2024-09-16
i18n: automatically extract the translations using a script

Msarawan:
The custom Yarn script "i18n" in the package.json facilitates the extraction of translation keys.

- Added script to automatically extract translation keys and values from the codebase.
- Implemented a CI check to ensure the build fails if strings are not properly extracted
or translations are missing.


lavocatt (approved):
LGTM, thanks for your work on this!

PR #291, opened 2024-09-17
[NO-ISSUE] fix: react unit test warning with 'list-style'

Msarawan:
- Updated the style property from 'list-style' to 'listStyle' to resolve unit test warning. 
- React requires inline styles to be written in camelCase rather than kebab-case.


lavocatt (approved):
Thanks a  lot for the fix!

Issue #294, opened 2024-09-18
Create a new workload category called Brokers


lavocatt:
ATM we host a list of brokers that is custom placed in a separate category. Since Brokers are workload, a shortcut should exist in the workload section. This new entry should display a list of brokers.

Current placement
![image](https://github.com/user-attachments/assets/b18a2160-e1a1-4624-9259-03048da17ace)

Desired placement, somewhere under workloads:
![image](https://github.com/user-attachments/assets/97827855-8bf0-4a2b-a5ad-90d1e0962e45)

What to do here:

- [x] Add the new entry point in the workload category
- [x] Make it link to the current list of brokers that is implemented when working on the issue (meaning that if #295 isn't done, it should be the custom brokers list instead)

PR #302, opened 2024-09-23
[#294] Added 'Brokers' entry within the Workloads section

Msarawan:
Since Brokers are a type of workload, a shortcut to our project has been added within the Workloads section as 'Brokers' which displays a list of brokers.

fixes: [#294](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/294)

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -60,22 +60,13 @@
     }
   },
   {
-    "type": "console.navigation/section",
-    "properties": {
-      "id": "activemq",
-      "perspective": "admin",
-      "name": "ActiveMQ Artemis",
-      "dataAttributes": { "data-quickstart-id": "qs-nav-home" }
-    }
-  },
-  {
     "type": "console.navigation/href",
     "properties": {
       "id": "brokers",
-      "name": "ActiveMQ Artemis Self Provisioning",
+      "name": "Brokers",
       "href": "/k8s/all-namespaces/brokers",
       "perspective": "admin",
-      "section": "activemq"
+      "section": "workloads"
     }
   }
 ]
--- a/README.md
+++ b/README.md
@@ -119,6 +119,8 @@ This will run the OpenShift console in a container connected to the cluster
 you've logged into. The plugin HTTP server runs on port 9001 with CORS enabled.
 Navigate to <http://localhost:9000> to see the running plugin.
 
+To view our plugin on OpenShift, navigate to the Workloads section. The plugin will be listed as **Brokers**.
+
 If you want the console to run in `https` mode, run:
 
 `yarn run start-console-tls`


console-extensions.json
```diff
@@ -77,5 +77,15 @@
...
       "section": "activemq"
     }
+  },
+  {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "brokers",
+      "name": "Brokers",
+      "href": "/k8s/all-namespaces/brokers",
+      "perspective": "admin",
```
lavocatt, line 87: I have a doubt on that part, why does it have to be limited to an admin perspective?

lavocatt (commented):
Thanks @Msarawan, that looks good. I have just a question down below.

lavocatt (changes requested):
When I'm selecting the brokers, I can see the proper page. However two sections are getting selected:

image

Can you update the contribution so that, when I click on broker, only this entry gets selected in the page?

Alternatively, if that isn't possible, let's completely remove the ActiveMQ Artemis entry. 

@RoddieKieley do you have an opinion on this?

Msarawan:
> When I'm selecting the brokers, I can see the proper page. However two sections are getting selected:
> 
> image
> 
> Can you update the contribution so that, when I click on `Brokers`, only this entry gets selected in the page?
> 
> Alternatively, if that isn't possible, let's completely remove the ActiveMQ Artemis entry.
> 
> @RoddieKieley do you have an opinion on this?

Hello @lavocatt , about the selection of both the entries i will take a look. but, we can't remove the entry with `Activemq Artemis` because in future we need to update this name with `Messaging` on openshift. so i think we cannot remove this entry as per our discussion atm. but will confirm from Roderick.

console-extensions.json
Msarawan: as far i know it's depends on our requirements. so initially we added `admin` perspective.

lavocatt:
Hey @Msarawan after checking with @RoddieKieley, the correct approach is to remove the `ActiveMQ Artemis` entry and keep only the `Brokers` one under `Workloads`

Msarawan:
> Hey @Msarawan after checking with @RoddieKieley, the correct approach is to remove the `ActiveMQ Artemis` entry and keep only the `Brokers` one under `Workloads`

I will update now, thanks!

console-extensions.json
```diff
@@ -59,23 +59,14 @@
...
-      "name": "ActiveMQ Artemis Self Provisioning",
+      "name": "Brokers",
       "href": "/k8s/all-namespaces/brokers",
       "perspective": "admin",
-      "section": "activemq"
+      "section": "workloads"
     }
   }
-]
+]
```
gaohoward, line 72: This removes the top-level plugin section. Maybe we can keep it along with `workloads` so people can quickly find the spp plugin? (maybe adding an `About` item)

console-extensions.json
```diff
@@ -59,23 +59,14 @@
...
   {
     "type": "console.navigation/href",
     "properties": {
       "id": "brokers",
-      "name": "ActiveMQ Artemis Self Provisioning",
+      "name": "Brokers",
       "href": "/k8s/all-namespaces/brokers",
       "perspective": "admin",
-      "section": "activemq"
+      "section": "workloads"
```
gaohoward, line 69: This removes the top-level plugin section. Maybe we can keep it along with workloads so people can quickly find the spp plugin? (maybe adding an `About` item  to show versions, branding, doc links...)

lavocatt: Yes that's a very good idea.

lavocatt: We've just had a meeting with @Msarawan  looking a the options for the about page, in patternfly, about pages seems to be modals and not actual page. We need to discuss the topic in today's UX meeting. I think in the meantime we can merge @Msarawan's work, if we need to bring back the menu it'll be easy to do anyways.

lavocatt (approved):
LGTM! Thanks

PR #307, opened 2024-09-30
fix: add missing spaces in the AcceptorConfigPage title

Msarawan:
Updated translation pattern to include missing spaces between the words in the AcceptorConfigPage title.

fixes: [#308](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/308)


lavocatt (approved):
LGTM! Thanks for the fix

PR #309, opened 2024-10-01
Integration of the custom broker edition form within OpenShift

Msarawan:
- Added integration of our custom broker-edition form within OpenShift. It allows users to edit brokers from multiple locations, such as through the OpenShift search results or from the Installed Operators section.

- Users can now utilize our custom broker-edition form to edit brokers in both instances.

Screenshot from 2024-10-01 13-54-58

Screenshot from 2024-10-01 13-55-20

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -18,10 +18,12 @@
   {
     "type": "console.page/route",
     "properties": {
-      "exact": true,
+      "exact": false,
       "path": [
         "/k8s/ns/:ns/edit-broker/:name",
-        "/k8s/all-namespaces/edit-broker/:name"
+        "/k8s/all-namespaces/edit-broker/:name",
+        "/k8s/ns/:ns/broker.amq.io~v1beta1~ActiveMQArtemis/:name",
+        "/k8s/ns/:ns/clusterserviceversions/:operator/broker.amq.io~v1beta1~ActiveMQArtemis/:name"
       ],
       "component": { "$codeRef": "UpdateBrokerContainer.UpdateBrokerPage" }
     }


console-extensions.json
```diff
@@ -18,10 +18,12 @@
...
     "type": "console.page/route",
     "properties": {
-      "exact": true,
+      "exact": false,
       "path": [
         "/k8s/ns/:ns/edit-broker/:name",
-        "/k8s/all-namespaces/edit-broker/:name"
+        "/k8s/all-namespaces/edit-broker/:name",
+        "/k8s/ns/:ns/broker.amq.io~v1beta1~ActiveMQArtemis/:name",
+        "/k8s/ns/:ns/clusterserviceversions/:operator/broker.amq.io~v1beta1~ActiveMQArtemis/:name"
```
lavocatt, line 26: While overriding the URLs with our custom component looks to be a good temporary solution, I looks to me it can't be a robust one. What if another version of the operator is installed?

This can be merged as addressing part of the issue, but I don't think we can consider that answering the completely the matter.

We need to ask the ui-extensibility folks if they want to add a new extension to the console like this one:
image

But for edition

Msarawan: Hello @lavocatt ,I have added :operator in the URL to dynamically retrieve the installed operator version, so it will automatically receive the new version whenever the operator is updated.

lavocatt: There's also this `v1Beta1` in the URL that is getting me slightly worried.

lavocatt: But that's the version of the `ActiveMQArtemis` so I guess that's fine.

Msarawan: Hello @lavocatt ,For the API version change, we need to update the logic in our component. Currently, adding the paths works perfectly fine, and if the API changes in the future, we will update our component to ensure it picks up the new version. Additionally, if the dynamic-plugin SDK provides an extension for this in the future, we will utilize that as well. wdys??

lavocatt: I've had a chat with the ui extensibility team, and they discourage overriding urls. I think we can merge this as a viable workaround but can't close the related issue. I've opened a RFE to have support for that feature https://issues.redhat.com/projects/RFE/issues/RFE-6377?filter=allissues

Once this is developed in the dynamic plugin sdk the related issue will get unblocked

lavocatt (approved):
That's a good workaround, let's wait up until the new extension reaches the dynamic plugin sdk to call the issue done, but in the meantime, good job!

Issue #305, opened 2024-09-26
Add a separator before `Brokers` in the workload 


lavocatt:
Using

![image](https://github.com/user-attachments/assets/b2885c24-575d-4d4c-acde-69fe798c7233)

Msarawan:
Hello @lavocatt , Thank you for suggesting the new `console.navigation/separator` extension. However, we can achieve the same result by simply adding the `insertAfter`and `insertBefore` properties to our existing `console.navigation/href` extension.

lavocatt:
Yes you are right, let's not add a separator, good finding.

PR #310, opened 2024-10-01
Move Brokers Entry After StatefulSets in the Workloads Section

Msarawan:
Reordered `Brokers` Entry to Follow StatefulSets in the Workloads Section.

Screenshot from 2024-10-01 14-13-44

fixes: [#305](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/issues/305)


lavocatt (approved):
Thanks, that's really better than before

PR #311, opened 2024-10-03
[NO-ISSUE] Revert the integration of the broker edition form within Openshift

Msarawan:
Reverted the integration of the custom broker edition form within OpenShift.


lavocatt (approved):
Thanks for the revert.

PR #339, opened 2024-12-05
Deployment should add plugin and remove on undeploy

michalxo:
This should make sure we won't accidentally remove user deployed plugins during un/deployment of our plugin.


lavocatt:
Can you add a description text with some context to your commit please?

gaohoward (approved):

lavocatt (approved):
Thanks @michalxo for this addition. This is working great and will avoid deleting other plugins in the console while we install ours.

Issue #340, opened 2024-12-10
Enable sending an identity token to the jolokia api server to identify the current user sending the requests.


lavocatt:
Following these documentations:

- https://github.com/openshift/console/tree/master/dynamic-demo-plugin#local-development-1
- https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

It's possible to wire a proxy service that will attach the user ID to every requests send to the service being proxied.

The goal is to have this working in dev & prod.

PR #341, opened 2024-12-11
[#340] WIP: proxy the jolokia api server

lavocatt:
In order to get authenticated as a user using the jolokia api server, we need to set up proxies. The console will take care of adding the authorization bearer token on every requests for us.

This commit adds the configuration for running both in dev mode and production mode.

### DIFF
--- a/bridge-auth-http/.gitignore
+++ b/bridge-auth-http/.gitignore
@@ -1,0 +1,3 @@
+ca.crt
+ca-bundle.crt
+console-client-secret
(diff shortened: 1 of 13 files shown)


gaohoward:
Suggestion: instead of using openshift template, consider using `kustomize` and put the resources under [deploy dir](https://github.com/artemiscloud/activemq-artemis-self-provisioning-plugin/tree/main/deploy)

